PR #758, opened 2012-09-02
Short filter syntax

coderanger:
This branch adds a new syntax of "tag::filter text" for single-line invocation of a filter.

My primary use case for this would be translation and localization tooling, but some others that come to mind include rendering math forumla and controlling text escaping.


tjconcept:
Is this in master? I can't find it anywhere in the docs. I'm considering it for i18n.

ForbesLindesay:
The pull request never got updated.  If someone wants to submit fresh pull requests against pug-lexer we could probably consider merging it.

TimothyGu:
It's already supported in current master (with a slightly different syntax):

```jade
tag: :markdown-it **bold**
```

```html
<tag><p><strong>bold</strong></p>
</tag>
```

ForbesLindesay:
I think the thinking here is that we probably shouldn't require that space in this particular case.

PR #1205, opened 2013-09-20
Add cli option to prevent adding any filetype extension

StephanHoyer:
I just want to compile a xml document called e.g. foo.xml.jade and it compiles it to foo.xml.html which is not very usefull.

So I added a cli-flag to indicate that no filetype extension should be added.


TimothyGu:
I have merged #1645 so this should be a non-issue.

PR #1255, opened 2013-10-17
Make filtering of include explicit

ForbesLindesay:
This adds the syntax:

``` jade
include:filter-name path/to/file.ext
```

It also means that if you had:

``` jade
include foo.js
```

before and expect it to carry on adding `<script>` tags, you will need to switch to either:

``` jade
include:js foo.js
```

or

``` jade
script
  include foo.js
```

### DIFF
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -377,6 +377,23 @@ Lexer.prototype = {
   },
 
   /**
+   * Include with filter
+   */
+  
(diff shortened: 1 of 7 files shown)


chharvey:
you could also just:

    script(src="foo.js")

MattyBalaam:
@chharvey that just adds links, not including the file though?

Issue #1261, opened 2013-10-21
Multiple blocks append in child layout block append and includes results in multiple bugs: blocks duplicated, with different orders


triccardi-systran:
The following examples are compiled using jade v0.35.0 and today's master (bd6233b61a408fd19c547442cac68e155da3dddc).

layout.jade
```jade
block scripts

block content
```

page2.jade
```jade
append scripts
  script console.log("page2");

append content
  p page2
```

page1-v1.jade
```jade
extends layout
append scripts
  script console.log("page1");

append content
  p page 1
```
page1-v1.html
```html
<script>console.log("page1");</script>
<p>page 1</p>
```
Up to here everything is OK.

page1-v2.jade
```jade
extends layout
append scripts
  script console.log("page1");

append content
  p page 1
  append scripts
    script console.log("page1 in content");
```
page1-v2.html
```html
<script>console.log("page1");</script>
<script>console.log("page1 in content");</script>
<p>page 1</p>
<script>console.log("page1 in content");</script>
<script>console.log("page1");</script>
```
__Issue 1__: block append in another block append duplicates the content of the inner block append, in different order furthermore.

page1-v3.jade
```jade
extends layout
append scripts
  script console.log("page1");

append content
  p page 1

include page 2
```
page1-v3.html
```html
<script>console.log("page2");</script>
<script>console.log("page1");</script>
<p>page 2</p>
<p>page 1</p>
```
__Issue 2__: append order is not the expected one: page 2 should be after page 1 in both script & content blocks.
With manual copy past of page2.jade instead of include page2.jade it works:
page1-v3bis.jade
```jade
extends layout
append scripts
  script console.log("page1");

append content
  p page 1

append scripts
  script console.log("page2");

append content
  p page 2
```
page1-v3bis.html
```html
<script>console.log("page1");</script>
<script>console.log("page2");</script>
<p>page 1</p>
<p>page 2</p>
```

page1-v4.jade (include inside append content)
```jade
extends layout
append scripts
  script console.log("page1");

append content
  p page 1
  include page2
```
page1-v4.html
```html
<script>console.log("page2");</script>
<script>console.log("page1");</script>
<p>page 1</p>
<script>console.log("page2");</script>
<script>console.log("page1");</script>
<p>page 2</p>
```
__Issue 3__: block scripts is duplicated like in __issue 1__, but the order is the same in the two copies...

This seems duplicated each time we append, directly or from an include:
page1-v4bis.jade (page 2 included two times + direct append)
[...]

Issue #997, opened 2013-04-28
include and block


cooolbasha:
I have a layout.jade

layout.jade

``` jade
block scripts

block content
```

i have page1.jade

``` jade
extends layout
append scripts
  :coffeescript
    console.log "page1"

append content
  p page 1
  include page2
  append scripts
    :coffeescript
        console.log "this work fine appended to the top of the page /script..but it is repeated with the previous script"
```

and page2.jade is

``` jade
append scripts
  :coffeescript 
     console.log "page 2"
append content
     p page2
```

the included page comes with the script at the end of page1  not at the top of the page1 along with other scripts..The script after the include page is added to to the top of page(and it is added once more again??). should not the page2 be included and then the extension be applied so that we can same behavior as .other append right after the include.

PR #1264, opened 2013-10-22
Add unit tests to describe bugs from #1261 and #997.

thomas-riccardi:
_include_.jade: the order of _block appends_ is not the expected one (file order)

other .jade: _block append_ should not return its value to its parent like _block_ does: it should only modify the block it's referring to.


thomas-riccardi:
Stopped working on this.

PR #1323, opened 2013-12-10
Support inline tags

### LINKED ISSUE #936: Feature Request: Nesting Syntax
So instead of having to do:

``` jade
   div
      | some text
      code some code
      | and some text
```

maybe we could do:

``` jade
   div some text [ code some code ] and some text
```

so it would treat everything inside the `[` `]` as it's own line then continue with the text node

not sure if this is a silly idea, thought id throw it out there

ForbesLindesay:
[closes #936]

This is the biggest roadblock in the way of a 1.0.0 release.  I'm not entirely happy with the implementation, but I think it's 90% of the way there, would love to get input from people.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -5,6 +5,7 @@ var nodes = require('./nodes');
 var utils = require('./utils');
 var filters = require('./filters');
 var path = require('path');
+var parseJSExpression = require('character-parser').parseMax;
 var extname = path.extname;
 
 /**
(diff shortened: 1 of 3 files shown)


caneta:
This is really nice and works well, but I cannot find it in the [documentation](https://pugjs.org/language/plain-text.html#inline-in-a-tag)...

TimothyGu:
@caneta, it is filed under [interpolation](https://pugjs.org/language/interpolation.html#tag-interpolation)

caneta:
I see...thanks @TimothyGu

PR #1392, opened 2014-01-21
Add a `lexer` option for the Parser

mquandalle:
Hello,

I'm currently building a jade package for the meteor framework. I need to do some modifications of the `Lexer` object and then ask the parser to use my custom object. Is that already possible? If no is that something that could be added?

Thank you


ForbesLindesay:
Support for overriding all stages of the pipeline will be added as part of 2.0.0

PR #1397, opened 2014-01-26
Adds a new `context` keyword.

timmfin:
It is similar to `include`, but instead of including the passed file's contents into the buffered output, the file's content is inserted directly into the compiled template code. This allows you to insert arbitrary javascript into the context of the template, such as JSON data, a bunch of helper javascript functions, etc.

This is particularly useful for things like static site generators that do not have some server middleware to pass in `locals` to the jade template.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -607,6 +607,17 @@ Compiler.prototype = {
   },
 
   /**
+   * Visit `context`
+   *
+   * @param {Context} context
+   * @api public
+   */
+
+  visitContext: function(context){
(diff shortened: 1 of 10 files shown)


ForbesLindesay:
Closing this in favour of the `import` proposal.  Thanks for all the hard work on this.  I promise I will land this feature in 2.0.0.

timmfin:
Looking forward to it, thanks!

NSExceptional:
@ForbesLindesay Did this ever get implemented? I don't see any documentation on it.

PR #1405, opened 2014-02-05
enhance stdin

yorkie:


TimothyGu:
To be honest, the correct way to do terminate input is through the "^D" escape code, which means "EOF". Allowing for "^C" is a hack.

vendethiel:
(this needs to be ^Z on windows)

TimothyGu:
@vendethiel true but still.

vendethiel:
I'm just saying — "if ^D goes through, please support ^Z"

yorkie:
hmm, i didn't call `^D` or `^C` directly, actually we are doing a software based nodejs, and the `^` stuffs is controlled by nodejs :-(

moreover, this case in my opinion doesn't mean the real EOF of the progress, user just says he wouldn't input anything, so the control is on jade/nodejs. and when inputing `^D`, it ends the whole progress/process, rather than input session.

TimothyGu:
@yorkie said:

> i didn't call `^D` or `^C` directly, and the `^` stuffs are controlled by nodejs :-(

It doesn't matter. `spawnedProc.stdin.emit('end')` Fast 'n' easy. Compared to `SIGINT` which you'll need to do `spawnedProc.kill('SIGINT')` with an unavoidingly ominous name.

> however when inputing `^D`, it ends the whole progress/process, rather than input session.

This is just plain wrong. `^C` means `SIGINT`, which ["terminal interrupt[s]"](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/signal.h.html) the program. `^C` *should* end the whole process.

On the other hand, `^D`/EOF/closing `stdin` is terminating the input stream, not the whole process.

PR #1407, opened 2014-02-06
Added the MIT license to the package.json.

keskival:
This is for automatic tools, for example: license-checker.
https://npmjs.org/package/license-checker


h5rdly:
Could you add a LICENSE file to github as well? This will allow to see that it's MIT right off the bat.

PR #1459, opened 2014-03-14
bring virtual fs support: decouple readFileSync from parser

ivankravchenko:
Bring virtual filesystem support to jade!
I've referred usage example in test case.


ForbesLindesay:
Thanks for all the work you've put into this.  I am going to land this feature in 2.0.0 but will do so slightly differently - by allowing the user to override the "loader" module in compilation.

Issue #1460, opened 2014-03-15
Inline tags and expressions aren't parsed


mquandalle:
Hello,

For my [jade implementation in Meteor](https://github.com/mquandalle/meteor-jade) I use the jade parser but a custom compiler. This approach works well except for inline tags and expressions were there are some inconsistencies.

Here are the parsed tree for each jade fragment:

``` jade
p
  | hello
  = name
```

``` json
{
  "nodes": [
    {
      "name": "p",
      "attrs": [],
      "attributeBlocks": [],
      "block": {
        "nodes": [
          {
            "val": "hello",
            "line": 2
          },
          {
            "val": "name",
            "buffer": true,
            "escape": true,
            "line": 3
          }
        ]
      },
      "selfClosing": false,
      "line": 1
    }
  ],
  "line": 0
}
```

---

``` jade
p hello #{name}
```

``` json
{
  "nodes": [
    {
      "name": "p",
      "attrs": [],
      "attributeBlocks": [],
      "block": {
        "nodes": [
          {
            "val": "hello #{name}",
            "line": 1
          }
        ]
      },
      "selfClosing": false,
      "line": 1
    }
  ],
  "line": 0
}
```

In the second case the `#{name}` isn't interpreted as a distinct leaf of the paragraph, but as a simple text value. That means that the compiler needs to [parse again](https://github.com/visionmedia/jade/blob/master/lib/compiler.js#L101-L103) this string of text.

Again for inline tags:

``` jade
p: strong hello world
```

``` json
{
  "nodes": [
    {
      "name": "p",
      "attrs": [],
      "attributeBlocks": [],
      "block": {
        "nodes": [
          {
            "name": "strong",
            "attrs": [],
            "attributeBlocks": [],
            "block": {
              "nodes": [
                {
                  "val": "hello world",
                  "line": 1
                }
              ]
            },
            "selfClosing": false,
            "line": 1
          }
        ]
      },
      "selfClosing": false,
      "line": 1
    }
  ],
  "line": 0
}
```

---

``` jade
p #strong[hello world]
```

``` json
{
  "nodes": [
    {
      "name": "p",
      "attrs": [],
      "attributeBlocks": [],
      "block": {
        "nodes": [
          {
            "val": "#strong[hello world]",
            "line": 1
          }
        ]
      },
      "selfClosing": false,
      "line": 1
    }
  ],
  "line": 0
}
```

Is there any reason for this behavior? Could we move the inline interpolation from the compiler to the parser?

PR #1461, opened 2014-03-15
Move inline expression parsing from the compiler to the parser

mquandalle:
See #1460.

Actually inline tags are parsed by the parser - as expected.
So the issue only concern inline code with the `#{}` syntax.


ForbesLindesay:
Thanks for all the work you've put into this.  I am completely overhauling the lexer/parser/compiler set up and will try and make sure that this gets looked at as part of that.

PR #1472, opened 2014-03-21
Use node-resolve to resolve paths if normal resolving fails

VanCoding:
I've extended the resolvePath function so, that it uses node-resolve as fallback if it cannot resolve the file.
This is especially useful if you want to include/extend from a view in a submodule.

I need this for one of my projects and of course it'd be cool to include this into the main repository.

What do you think?

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -128,6 +128,13 @@ Parser.prototype = {
       this.context(parser);
       var ast = parser.parse();
       this.context();
+      
+      //append main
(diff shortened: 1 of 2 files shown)


ForbesLindesay:
Thanks for all the work that went into this.  I definitely plan to land something along these lines in 2.0.0.  My idea is to implement `import foo from 'bar'` support in the same way as ES6 does, except it will also support importing mixins from npm modules.

PR #1523, opened 2014-05-10
Optional module.exports syntax

bevacqua:
When using Browserify or similar CJS solutions for the client-side, it's nice having the ability to have `jade` create modules rather than many `function template () { ... }` files

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ $ jade --client --no-debug filename.jade
 
 which will produce `filename.js` containing the compiled template.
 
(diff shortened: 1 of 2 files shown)


shanemikel:
This would allow me to use require('compiled-jade.js')(locals) to serve halfway-precompiled templates from node.  Anybody have an idea if this is much faster than on-the-fly server-side rendering?

bevacqua:
You can use [jadum](https://github.com/bevacqua/jadum) for that

shanemikel:
thanks @bevacqua

PR #1525, opened 2014-05-12
export to correct directory even if sub-directories

bevacqua:
Without this change, if you use an `out` directory and your templates are in sub-directories, the directory structure gets squished. Think you had controllers and views per controller, then that hierarchy gets flattened.

This patch fixes that issue.


ForbesLindesay:
Closing in favour of #1829

PR #1556, opened 2014-06-14
mixin precompilation

jeromew:
Hello,

This PR is hopefully an acceptable way to separate the compilation of the mixins from the compilation of the templates.

In other words, you can compile mixins once (as a bundle or one by one) and then the compilation of a template will not need to recompile all the mixins :-)

I have used it for some time now and did not run into issues. All tests are passing.

I think this might be a nice addition for the dynamic capabilities of jade 
 - combined with the dynamic mixin calls
 - to allow for easier componentization of mixins (as npm modules for instance)

Tell me what you think.

Example code:

mixin.jade
````
mixin dynamic(world)
 | hello #{world}
````

 template.jade
````
+dynamic("world")
````

index.js
````
var jade = require('jade');
var locals = {
  keepMixins: true,
  jade_mixins: {}
};

var mixin = jade.renderFile('mixin.jade', locals);
var html = jade.renderFile('template.jade', locals);
console.log(html);
````

=> result
````
hello world
````

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -40,7 +40,7 @@ var Compiler = module.exports = function Compiler(node, options) {
   this.parentIndents = 0;
   this.terse = false;
   this.mixins = {};
-  this.dynamicMixins = false;
+  this.dynamicMixins = options.keepMixins ? true : false;
   if (options.doctype) this.setDoctype(options.doctype);
 };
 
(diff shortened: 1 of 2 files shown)


ForbesLindesay:
I plan to support something like this as part of 2.0.0 via some kind of `import foo from 'bar'` syntax along the same lines as ES6 modules.  Thanks for all the work that you've put into this.

PR #1636, opened 2014-08-22
Space control

char101:
Hi,

This is a proposal to add space before and after the tag.

```jade
p
  span no space
p
  _span space before
p
  span_ space after
p
  _span_ space before and after
```

```html
<p><span>no space</span></p><p> <span>space before</span></p><p><span>space after</span> </p><p> <span>space before and after</span> </p>
```

```jade
p
  | This is a
  _a_(href="#") link
  | one
```

```html
<p>This is a <a href="#">link</a> one</p>
```

```jade
.btn-group
  button.btn Button 1
  _button.btn Give me some space
```

```html
<div class="btn-group"><button class="btn">Button 1</button> <button class="btn">Give me some space</button></div>
```

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -439,6 +439,7 @@ Compiler.prototype = {
     if (pp && !tag.isInline())
       this.prettyIndent(0, true);
 
+    this.buffer(tag.before);
     if (tag.selfClosing || (!this.xml && selfClosing.indexOf(tag.name) !== -1)) {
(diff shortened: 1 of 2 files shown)


ForbesLindesay:
I will disallow tags that start or end with an `_` as part of 2.0.0.  I don't currently want to add this feature, but by making it a syntax error, we will be giving ourselves the option of adding it as a non-breaking change in the future.

PR #1642, opened 2014-08-30
Create a new option, `options.loadPaths`

timmfin:
This is similar to `options.filename` and `options.basedir`, but is an array. It allows you to tell jade to search for include/exclude paths in several possible directories (instead of just one).

Something like this is helpful when trying to share groups of jade files (jade modules?) among several different projects. That way, every project that uses these shared modules can refer to them by the same path. For example:

``` jade
//- Ideally several projects would refer to these shared jade files like
include module1/sometmpl 
extend module2/anothertmpl

//- And not have to do these things like this differently in each project, depending on how exactly that “jade module” was included in that project
include ../../lib/module1/sometmpl 
extend /vendor/assets/jade/module2/anothertmpl
```

This gives jade includes/extends a _lot_ more flexibility. Especially when used in a non-node context (like the Rails asset pipeline or some static site generator).

Some similar precedent/examples are [load_paths in SASS](http://sass-lang.com/documentation/Sass.html#load_paths-class_method), the [paths option in Stylus](http://learnboost.github.io/stylus/docs/import.html#stylus-import), and the NODE_PATH environment variable for node.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -504,8 +504,7 @@ Parser.prototype = {
 
   searchForPath: function(partialPath, dirsToCheck, jadeExtensionRequired) {
     var fs = require('fs'),
-        p = require('path'),
-        validPath;
+        p = require('path');
 
     if (jadeExtensionRequired === undefined) {
       jadeExtensionRequired = false;
@@ -528,8 +527,6 @@ Parser.prototype = {
         return pathToCheck;
       }
     }
-
-    return validPath;
   },
 
   /**


vancarney:
I think it's worth pointing out that recent versions of Express (>=4.12) now supports passing in an array of view directories as follows:
```
// set .jade as the default extension
app.set('view engine', 'jade');

// define view dirs as Array
app.set('views', ['path/to/view-set-one','path/to/view-set-two');
```

However neither EJS or Jade support this internally, so all partials/includes/inherits/extends must be located in the same base path the template was discovered in.

It's my opinion that the lack of support for multi-paths within the engines themselves makes this feature of Express rather weak. Ideally, these view paths could be passed to the template engine which could then do multi-directory searches for each template include.

Perhaps it is time to revisit this issue as I think it is useful, is common in other platforms as such Ruby and PHP and when coupled with caching, can be optimized for production use without the overhead of searching for various templates on each request. 

Most importantly, supporting this feature would provide the expected behavior that Express' multi-path support entails.

Issue #1633, opened 2014-08-20
Evolution of mixins - proposal / draft


Artazor:
Hi @ForbesLindesay and jade-community!
I would like to propose or discuss the impact (features vs. performance)  of the following refactoring (and new features). Proposed refactoring is conservative in terms of public api - all existing unit tests are passed. 
## TL;DR
1. Let's make templates and mixins first-class values in Jade in a clean sound and backward-compatible way:
   
   ```
   +[expression_that_evaluates_to_mixin](1,2,3)#id.some-class(href="#")
       span Hello from block
   ```
2. And let's make templates and mixins more OOP friendly by allowing them to be late-bound on the same prototypal resolution principles as method calls:
   
   ```
   h1= this.title
   ul: each book in this.books
       li: .book
           +[book]cover()
           +[book.author]#{"bio" + "graphy"}()
   ```
## TL;DR-2 (longer)

_Since subclassing is effectively constructed by prototype chaining, we can actually speak about classes, inheritance and overriding._

Imagine that you can _somehow_ attach templates to your classes and use `this` reference to access current instance for accessing properties, calling methods and rendering other _member mixins_. 
##### class Person

``` javascript
function Person(first_name, last_name, photo_url) {
     this.first_name = first_name;
     this.last_name = last_name;
     this.photo_url = photo_url;
}

var person_mixins = jade.mixinsFor(Person);

person_mixins["display-name"] = jade.compileMixin(src1);
person_mixins["info"] = jade.compileMixin(src2);
```

where _src1_ and _src2_ are:

``` jade
//- mixin Person::display-name (src1)
span.display-name
    span.first-name= this.first_name
    span.last-name= this.last_name
```

``` jade
//- mixin Person::info (src2)
.person
     img.photo(src=this.photo_url)
     +[this]display-name
```
##### class Agent

``` javascript
function Agent(code_name, photo_url) {
     this.code_name = code_name;
     this.photo_url = photo_url;
}

var agent_mixins = jade.mixinsFor(Agent);

agent_mixins["info"] = jade.compileMixin(src3);
```

where _src3_ is:

``` jade
//- mixin Agent::info (src3)
.agent
     img.photo(src=this.photo_url)
     span.code-name= this.code_name
```
##### Mixin late binding demo

For given array of objects

``` javascript
var persons = [ 
      new Person("Thomas","Anderson","/img/neo.jpg"),
      new Agent("Smith","/img/smith.jpg")
  ];
```

the following code

``` jade
[...]

Issue #1608, opened 2014-07-28
Callable block-mixin


KidsKilla:
It'd be great to use custom mixins as "callable block" like iterators:
```jade
- var dbg = {a:1, b:2, c:3}
mixin eachDbg()
    ul
        each val, name in dbg
            +block(name, val, name + val)
foo
    +eachDbg
        mixin (name, val ,id)
            tab: a(href="##{id}")= name
    //
    +eachDbg
        mixin (name, val, id)
            h2= name
            pane(id=id)= val
```
I hope this example explains my idea well

ForbesLindesay:
Mixins are being overhauled as part of 2.0.0 so hopefully this will be resolved by that.

PR #1654, opened 2014-09-05
First-class mixins backward-compatible (non-breaking)

Artazor:
Hi! @ForbesLindesay & @visionmedia 
This is rather big commit that can't be easily splitted into progressive atomic commits, since many interconnected parts are refactored (very carefuly with respect to  the current architecture of Jade)

I'll describe all changes in detail later, however here are most notable news:

1. First-class mixins now supported. They have backward compatible signatures, `buf` is now passed in the same way as `block` and `attributes` via `.call({...}, args)`. If no `this` context provided for call, all such values are inherited from outer scope. The same is true for `self` reference. However all special local variables are present in scope (buf, jade, block, attributes) and work consistently.
2. `mixin (args)` syntax now produces mixin literal (anonymous mixin). Mixin literal can be used as block for other mixin call, or placed into separate file and compiled with `jade.compileMixin` or `jade.compileMixinFile`.
3. `+[mixinExpression](args).my-class(href="#")`  syntax can be used for calling mixin-values. 
4. `+[obj]name(args...)` is equivalent to 

        - jade.resolveMixin(obj,"name").call({buf:buf,self:obj,...},args...)

     and interpolation of name `+[obj]#{nameExpr}(args...)` is equivalent to 

        - jade.resolveMixin(obj,nameExpr).call({buf:buf,self:obj,...},args...)

    where `jade.resolveMixin` can be used for late-bound mixin dispatching or even loading them on the fly.
5. Now mixins are properly scoped and can be nested without fear of redefinition
6. All this functionality is available because of brand new method `runtime.spawn` (that based on prototyping) [see the code]
7. Instead of `nodes.Mixin` with different flags now there are three distinct node types in AST: `nodes.Mixin`, `nodes.MixinLiteral` and `nodes.Call` generated by appropriate `parser.parseXXX` and compiled with `compiler.visitXXX`
8. Principle of least surprise is everywhere met. Code refactored very carfully, with respect to current design, no new hacks introduced, but some old ones removed because of proper decomposition.
9. All existing unit-tests are passed in their original state.
10. Unit tests that will demonstate new syntax is on the way (on weekend).

It would be great if somebody from core-team could take a look on this pull request. Diff is not comprehensive, so it's better to look at the resulting code directly, in particular how new nodes are compiled. And the trick with `runtime.spawn` that heavily based on prototyping.

It would be great to join to the jade-team and discuss this (or further) work in depth.
Best regards,
Anatoly. 

P.S. This PR Is closely related to #1608, #1633 and especially #1556 - and hopefully solves mentioned problems in non-breaking fashion.


ForbesLindesay:
I will be overhauling mixins as part of 2.0.0 and will take these ideas into consideration.  For now I'm going to close this pull request as I'm not sure it's quite how we're going to go about this.

PR #1673, opened 2014-09-17
Disallow colon at start or end of attributes

### LINKED ISSUE #1182: Jade should not let us generate invalid HTML
Occasionally my co-lunatics and I drink too much coffee and start writing Jade like this:

```
console.log("Things are going completely swimmingly");

option(value="level1", class : selected)
    | Level One
```

Naturally this turns into the sort of HTML not even my grandmother would parse:

```
<console Things are going completely swimmingly class="log">;</console>

<option value="level1" class : selected>Level One</option>
```

But Jade quite happily sends this misleading gibberish across the Internet, whistling nonchalantly, but with a curl at the edge of its mouth and an evil glint in its eye.

Since we are having trouble kicking the habit, I wonder if Jade might be able to help us out.  Could Jade possibly define what is **valid Jade**, and what is **valid HTML**, and warn us if we exceed the reasonable specifications?  Of course this could be made optional, just in case W3C switches to the same brand of coffee in future, or for those people breeding their own mutant army of non-compliant browsers.

Or is the above behaviour entirely reasonable for an HTML templating engine?  (I feel both lines should be invalid Jade, and the second line is non-compliant HTML.)  Perhaps instead you would recommend some validation/linting middleware that can throw warnings at us during development?

Thanks!  I hope this doesn't distract you from ensuring comments parse correctly!  ;)

ForbesLindesay:
[closes #1182]

This prevents things like:

```jade
option(value="level1", class: selected)
```

from compiling, but if you really wanted the old behavior, you just have to quote the attribute:

```jade
option(value="level1", "class:" selected)
```

```html
<option value="level1" class:="class:" selected="selected">Level One</option>
```

This is a breaking change, so part of the 2.0.0 track.

### DIFF
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -668,6 +668,9 @@ Lexer.prototype = {
           if (val) assertExpression(val)
           key = key.trim();
           key = key.replace(/^['"]|['"]$/g, '');
+          if (key[0] === ':' || key[key.length - 1] === ':') {
+            throw new Error('":" is not valid as the start or end of an un-quoted attribute.');
+          }
           tok.attrs.push({
             name: key,
             val: '' == val ? true : val,
--- a/test/anti-cases/colon-attribute-2.jade
+++ b/test/anti-cases/colon-attribute-2.jade
@@ -1,0 +1,1 @@
+option(value="level1", class: selected)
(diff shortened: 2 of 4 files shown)


TimothyGu:
This is no longer relevant in this repo; moving to jadejs/jade-lexer#21.

PR #1737, opened 2014-11-17
Warn for possible changes on the `context` command

TimothyGu:
See #1397.

### DIFF
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -227,6 +227,12 @@ Lexer.prototype = {
       } else {
         tok = this.tok('tag', name);
       }
+      if (tok.val === 'context') {
+        console.warn('Using `context` as an HTML tag is deprecated' +
+            ' and `context` will have a differnet meaning in Jade 2.0.0,' +
+            ' for line ' + this.lineno +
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
We don't necessarily have to implement it using require (although I tend towards that as a strong possibility.

danielbayley:
So what was the result of this? I arrived here from #1397 looking for the _proper_ way to do the same thing. Is it `require` and ES6 modules?

I realise the enforced move to _Pug_ has left things in a bit of a transitional mess, but the documentation is very unclear.

TimothyGu:
> So what was the result of this?

We haven't implemented this yet. As a result, unfortunately if you want this feature you have to make this imported file look like

```jade
-
  var whateverYouWantToPutHere = 0;
  var anotherVariable = {};
```

and `include` this file.

> I realise the enforced move to Pug has left things in a bit of a transitional mess, but the documentation is very unclear.

That's one of the reasons why we left pug in alpha right now. Most documentation has been written but not deployed yet for this reason.

tycho01:
Might there be any open thread we could subscribe to to be informed in the event of further progress?

PR #1802, opened 2014-12-31
Revert "Updated License."

TimothyGu:
This reverts commit b62bc4605619fdc062a8e63845a3c1451f255d9f.

1. It's not 2015 here yet, though it soon will be.
2. TJ didn't do anything in 2015, so the copyright year would be wrong.

CC @hemanth

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 (The MIT License)
 
-Copyright (c) 2009-2015 TJ Holowaychuk <tj@vision-media.ca>
+Copyright (c) 2009-2014 TJ Holowaychuk <tj@vision-media.ca>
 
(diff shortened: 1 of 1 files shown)


hemanth:
@TimothyGu Yeah, noticed that in `koa`. Let's revise this soon.

ForbesLindesay:
Do we now need to update the year to 2015 since it actually is 2015 now?

vendethiel:
NO. It's completly pointless to change a date like that. Could even be bad (because you're "cheating the system").

ForbesLindesay:
Ok, I just have virtually no understanding of the licensing stuff.  Especially with contributors in multiple countries :)

vendethiel:
Right. For example, for me, in France, there is no such thing as a "copyright".

Issue #1795, opened 2014-12-23
CLI option `jade --watch` broken 


flintinatux:
When using the `--watch` option in CLI mode, the first render works fine, but subsequent renders fail on file changes with an error message similar to the following.

```
$ jade -P -w client/*.jade -o public

\
  rendered public/index.html
TypeError: path must be a string
    at Object.fs.lstat (fs.js:669:11)
    at renderFile (/Users/scott/flintinatux/mithril-todos/node_modules/jade/bin/jade.js:172:6)
    at StatWatcher.<anonymous> (/Users/scott/flintinatux/mithril-todos/node_modules/jade/bin/jade.js:119:11)
    at StatWatcher.emit (events.js:98:17)
    at StatWatcher._handle.onchange (fs.js:1109:10)
```

This is a not an issue with the previous version (1.8.1), so for now I've just locked my dependency to that version.

TimothyGu:
@flintinatux See #1803 for a rudimentary test for the watch mode. I would appreciate any reviews as I am nervous about all the spawning trickery/hackery.

robmuh:
Just want to thank you for your quick attention to this. I've confirmed it is working as well. Will this be in a 1.8.3 release eventually? No rush, but semester starts on Jan 5th and my Web-2 students usually get thrown into Jade from the very beginning. (We start with basic `npm install -g jade` and then `-w` stuff instead of the other prepros tools and workflows to get a feel for what is happening.)

TimothyGu:
@robmuh I am sure it will be in 1.8.3, but the release date for 1.8.3 depends entirely on the availability of @ForbesLindesay. BTW, there is a college course for Jade? I would really want to take that class.

robmuh:
Thanks for the info. Not college, our nano-school (SkilStak Coding Arts)[http://skilstak.com) where I've been introducing kids as young as 10 to Jade after they get a general feel for HTML/CSS, etc. Jade and Stylus actually make web development easier for these guys because they have started their code lives on Python.

TimothyGu:
@robmuh Ah ok. I'm just a little bit too old. :)

PR #1803, opened 2015-01-01
Some more tests

TimothyGu:
This PR could really use some more eyes, especially for the watch mode test. I'm kind of anxious of all the child_process hackery/tricks as I'm not that proficient in Node.js.

See also #1795.

### DIFF
--- a/test/command-line.js
+++ b/test/command-line.js
@@ -3,10 +3,10 @@
 var fs = require('fs');
 var path = require('path');
 var assert = require('assert');
-var exec = require('child_process').exec;
+var cp = require('child_process');
 
 function run(args, callback) {
-  exec('node ' + JSON.stringify(path.resolve(__dirname + '/../bin/jade.js')) + ' ' + args, {
+  cp.exec('node ' + JSON.stringify(path.resolve(__dirname + '/../bin/jade.js')) + ' ' + args, {
     cwd: __dirname + '/temp'
   }, callback);
 }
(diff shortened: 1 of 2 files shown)


test/command-line.js
```diff
@@ -40,4 +40,49 @@ describe('command line', function () {
...
+
+    watchProc
+      .on('error', done)
+      .stdout.setEncoding('utf8');
+    watchProc.stdout.on('data', function(buf) {
+        stdout += buf;
+        if (/.*rendered.*/.test(stdout)) {
+          stdout = '';
+          var template = Function('', fs.readFileSync(__dirname + '/temp/input-file.js', 'utf8') + ';return inputFileTemplate;')();
+          console.log(template())
```
flintinatux: Should probably remove any logging before you merge.  Funny thing is: I noticed it in the Travis CI output before I spotted it here!

TimothyGu: @flintinatux oops yes :)

test/command-line.js
```diff
@@ -40,4 +40,49 @@ describe('command line', function () {
...
+  it('jade --no-debug --client --name-after-file --watch input-file.jade (pass 2)', function (done) {
+    fs.writeFileSync(__dirname + '/temp/input-file.js', 'throw new Error("output not written (pass 2)");');
+    fs.writeFileSync(__dirname + '/temp/input-file.jade', '.foo baz');
+
+    watchProc
+      .on('error', done)
+      .stdout.on('data', function(data) {
+        stdout += data;
+        if (/.*rendered.*/.test(stdout)) {
+          watchProc.kill('SIGINT');
```
flintinatux: I really like the way this is all done.  The asynchronous nature of file watching is difficult to capture, but I think you nailed it.

My only concern is that if the test times out for some reason, the `watchProc` won't get killed since that only happens in this listener.  I'm thinking we should move it out into an `after` [hook](http://mochajs.org/#hooks) to prevent any zombies.  Something like:
```javascript
describe('jade --watch', function() {  // enclose the whole suite
  var watchProc;
  var stdout = '';
  it('watch test one', function() {})  // first test
  it('watch test two', function() {})  // second test

  after(function() {  // runs after the suite (both tests) finish
    if (watchProc != null) watchProc.kill('SIGINT');
  });
});
```

flintinatux: Although now that I'm rereading the mocha docs, you may need to move that `after` hook to before the `it` examples.

TimothyGu: OK, that sounds a lot better than my current hack. Will try to implement it.

TimothyGu:
@flintinatux Done.

test/command-line.js
```diff
@@ -41,3 +41,54 @@ describe('command line', function () {
...
+  it('jade --no-debug --client --name-after-file --watch input-file.jade (pass 1)', function (done) {
+    fs.writeFileSync(__dirname + '/temp/input-file.jade', '.foo bar');
+    fs.writeFileSync(__dirname + '/temp/input-file.js', 'throw new Error("output not written (pass 1)");');
+    var args = [__dirname + '/../bin/jade.js', '--no-debug', '--client', '--name-after-file', '--watch', 'input-file.jade']
+    watchProc = cp.spawn('node', args,  {
+      cwd: __dirname + '/temp'
+    });
+
+    watchProc.stdout.setEncoding('utf8');
+    watchProc.on('error', function(err) {
```
flintinatux: Did passing in `done` as the callback not work here?  I thought it was much cleaner that way.

test/command-line.js
```diff
@@ -41,3 +41,54 @@ describe('command line', function () {
...
+        watchProc.removeAllListeners('error');
+        return done();
+      }
+    })
+  });
+  it('jade --no-debug --client --name-after-file --watch input-file.jade (pass 2)', function (done) {
+    fs.writeFileSync(__dirname + '/temp/input-file.js', 'throw new Error("output not written (pass 2)");');
+    fs.writeFileSync(__dirname + '/temp/input-file.jade', '.foo baz');
+
+    watchProc.on('error', function(err) {
```
flintinatux: Same here.

test/command-line.js
TimothyGu: I thought it works better, but it apparently doesn't. Removed.

TimothyGu:
@flintinatux change reverted

flintinatux:
Looks good to me.  I say ship it!  Thanks for all the work you put into this.

TimothyGu:
@flintinatux Merged. Thanks for the review!

PR #1808, opened 2015-01-03
Caching and caching tests improvements

TimothyGu:


TimothyGu:
I'll merge this in two days if no one reviews, to keep PRs from rotting.

ForbesLindesay:
If we did,`handleSrcCache` as:

```js
function handleFileCache (options) {
  if (options.cache && exports.cache[options.filename]) {
    return exports.cache[options.filename];
  } else {
    var result = exports.compile(fs.readFileSync(options.filename, 'utf8'), options);
    if (cache) exports.cache[options.filename] = result;
    return result;
  }
}
```

We could skip the extra (totally pointless) cache of the source strings.

vendethiel:
(2nd if's condition should read `options.cache`)

ForbesLindesay:
Thanks, updated

TimothyGu:
IIRC I tried that but it didn't work for some obscure reason I do not recall right now. Will get back to you.

TimothyGu:
Updated PR. Removed handleSrcCache() altogether.

lib/index.js
```diff
@@ -343,13 +356,8 @@ exports.renderFile = function(path, options, fn){
...
   options = options || {};
 
-  var key = path + ':string';
-
   options.filename = path;
-  var str = options.cache
-    ? exports.cache[key] || (exports.cache[key] = fs.readFileSync(path, 'utf8'))
-    : fs.readFileSync(path, 'utf8');
-  return exports.render(str, options);
+  return exports.render(null, options);
```
ForbesLindesay: Maybe it would be better if this did:

```js
return handleTemplateCache(options)(options);
```

It's not much duplicated code, and it keeps `exports.render` from needing to accept `null` for the string.  We could then add a check in `exports.render` that throws if `str === null` to catch that odd corner case.  I'd hate to end up having to support it forever.

TimothyGu: True.

PR #1818, opened 2015-01-12
No process uncaught exception

ForbesLindesay:
This pull request contains 3 key changes:

1. This makes the CLI use synchronous code.  I very much doubt there is any real performance impact given that the jade rendering is the slow part and that was synchronous anyway.  It vastly simplifies the code, and in particular makes it way easier to do `--watch` mode.
2. It makes `--watch` mode use a `try/catch` block rather than `process.on('uncaughtException', ...)` (see the [node docs](http://nodejs.org/api/process.html#process_event_uncaughtexception) for why `process.on('uncaughtException', ...)` is a really bad idea here).
3. It adds a 1 second delay between each of the `--watch` tests.  This was necessary to get the tests passing on my mac (with or without the other two changes).  This is because the `mtime` of the files is only to the nearest second, so it won't rebuild files that change twice in a single second.  My guess is that the travis machine was just slow enough to get away with it.

Once this is merged, I will cut a new release as we have some significant bug fixes, but I don't want to release another version with `process.on('uncaughtException')`.

### DIFF
--- a/bin/jade.js
+++ b/bin/jade.js
@@ -117,7 +117,7 @@ if (files.length) {
       }
       fs.watchFile(filename, {persistent: true, interval: 200},
                    function (curr, prev) {
-        if (curr.mtime === prev.mtime) return;
+        if (curr.mtime.getTime() === prev.mtime.getTime()) return;
         try {
           renderFile(filename);
         } catch (e) {
--- a/bin/jade.js
+++ b/bin/jade.js
@@ -109,16 +109,21 @@ if (files.length) {
   console.log();
   if (options.watch) {
     files.forEach(function(filename) {
-      // keep watching when error occured.
-      process.on('uncaughtException', function(err) {
-        console.error(err.stack
-                      || /* istanbul ignore next */ (err.message || err));
-      });
-      renderFile(filename);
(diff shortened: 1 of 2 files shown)


TimothyGu:
Thanks for fixing this. Didn't realize that throw-catch didn't work on async functions. (I'm still relative new to JavaScript and Node.js.)

TimothyGu:
> This is because the `mtime` of the files is only to the nearest second, so it won't rebuild files that change twice in a single second. My guess is that the travis machine was just slow enough to get away with it.

Probably. And my machine too ;)

> Once this is merged, I will cut a new release as we have some significant bug fixes, but I don't want to release another version with `process.on('uncaughtException')`.

Cool. Sorry for not sending a PR in the first place.

ForbesLindesay:
Yes, error handling with callbacks, hence [promises](https://www.promisejs.org/) which I would use here, but we really don't need async anyway so it's more effort than it's worth.

ForbesLindesay:
I'll merge this in the next hour or so if nobody objects so we can get a release out.

TimothyGu:
LGTM.

PR #1820, opened 2015-01-13
Add Coveralls support

TimothyGu:
See https://coveralls.io/r/TimothyGu/jade?branch=coveralls for an example


ForbesLindesay:
I'd like to put:

```
cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
```

in an npm script as:

```
"coveralls": "cat ./coverage/lcov.info | coveralls"
```

then just put `npm run coveralls` in the travis-ci script, but other than that it looks great.

Another thing that may be worth thinking about: we could remove coverage from the default `npm test` script?  It does slow down testing quite a lot when you're not actively looking for improvements to test coverage.

TimothyGu:
@ForbesLindesay said:
> I'd like to put:
>
> ```
> cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
> ```
>
> in an npm script as:
>
> ```
> "coveralls": "cat ./coverage/lcov.info | coveralls"
> ```
>
> then just put `npm run coveralls` in the travis-ci script, but other than that it looks great.

It's fine for me, but one thing to think about is a developer is unlikely to actually run the npm command locally, as it depends on some Coveralls tokens that even I do not have access to. Travis CI works for some reason.

> Another thing that may be worth thinking about: we could remove coverage from the default `npm test` script?  It does slow down testing quite a lot when you're not actively looking for improvements to test coverage.

I would like to +1 this. The sole reason why `npm test` includes `npm run coverage` is probably because Travis runs `test` by default, and it would be better to make sure istanbul runs correctly as well. But now we are adding a script in .travis.yml anyway so I think this is a good time to change it.

ForbesLindesay:
All sounds good to me.  I think I still prefer to put it in package.json personally, purely because that file path looks brittle to me.  I assume people will get a sensible error if they try and run it.

TimothyGu:
OK, well then. The exact error message printed is:

> ```
> Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
> ```

which IMO is probably enough to keep users from trying to do that ;)

TimothyGu:
Done. Merging.

PR #1821, opened 2015-01-13
Fix examples

### LINKED ISSUE #1819: Fix all the examples
Many examples don't work. The example test doesn't even work. It just tries to `require` it and exit normally it it breaks. That's not good.

TODO:
- [ ] Fix the test
- [x] csrf.js
- [x] dynamicscript.js

TimothyGu:
Fixes #1819.


TimothyGu:
Got a test working too. Will push in a few minutes/

examples/csrf.js
```diff
@@ -8,35 +8,36 @@ var jade = require('../'),
...
+    for (var i = 0; i < node.attrs.length; i++) {
+      var attr = node.attrs[i];
+      if (attr && attr.name === 'method' && attr.val === "'post'") {
         var tok = new nodes.Tag('input');
-        tok.setAttribute('type', '"hidden"');
-        tok.setAttribute('name', '"csrf"');
-        tok.setAttribute('value', 'csrf');
+        tok.setAttribute('type', '"hidden"', true);
+        tok.setAttribute('name', '"csrf"', true);
+        tok.setAttribute('value', 'csrf', false);
```
ForbesLindesay: This API should be considered deprecated.  I'm planning to completely overhall the AST as part of the 2.0.0 rewrite.  We should probably just remove this example.

I would like to support plugins properly afterwards though, and as part of that I'd like to consider the AST format part of the public API.

It might be worth doing something to add a deprecation notice to `.setAttribute`, but we should be careful to make sure it doesn't get output when jade uses the method internally.

TimothyGu: There is a warning in `getAttribute` so that's why I reimplemented it here. I'm fine with removing it, as nobody probably uses it or even understands it anyway (which explains why it's broken).

ForbesLindesay: Yes, `getAttribute` was consciously deprecated because we don't use it inernally, but `setAttribute` wasn't as easy to deprecate.  We will be removing it in the future.

P.S.  I actually really like the idea behind this plugin (automatically adding a csrf token to all forms) so I'd love to have that as an example of the new plugin API.  Preferably available to install from npm as well.

TimothyGu: OK, so do you want me to remove this example for now?

ForbesLindesay: Yes, I think that's probably the best bet.

PR #1822, opened 2015-01-13
Cleanup coverage

ForbesLindesay:
This splits the coverage task out by adding “precoverage” and “postcoverage”.  It also deletes the intermediate coverage files once it is done.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,9 @@
   },
   "scripts": {
     "test": "mocha -R spec && npm run coverage",
-    "coverage": "rimraf cov-pt* && istanbul cover --report none --dir cov-pt0 node_modules/mocha/bin/_mocha && istanbul report --include cov-pt\\*/coverage.json",
+    "precoverage": "rimraf coverage && rimraf cov-pt*",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -66,7 +66,9 @@
   },
   "scripts": {
     "test": "mocha -R spec && npm run coverage",
-    "coverage": "rimraf cov-pt* && istanbul cover --report none --dir cov-pt0 node_modules/mocha/bin/_mocha && istanbul report --include cov-pt\\*/coverage.json",
+    "precoverage": "rimraf coverage && rimraf cov-pt*",
+    "coverage": "istanbul cover --report none --dir cov-pt0 node_modules/mocha/bin/_mocha -- -R dot",
```
TimothyGu: Why `-R dot`?

TimothyGu:
Are you sure the `pre` and `post` work as expected?

package.json
TimothyGu: Actually I get why you put it there now. We already have a dedicated `test` script with proper output.

ForbesLindesay:
Yes, "pre" and "post" work for any command (unless that command already starts with "pre" or "post", it's not recursive).  They make for a really nice clean way to do setup and cleanup :)

PR #1824, opened 2015-01-13
Do not run coverage in `npm test`

TimothyGu:
Running coverage report is never the point of `test`.

See also https://github.com/jadejs/jade/pull/1820#issuecomment-69693312.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     }
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


TimothyGu:
This reduces the Travis run time to pre-#1820 level.

ForbesLindesay:
Yep, looks good to me

PR #1827, opened 2015-01-15
Do not try to read a file if src string is defined

### LINKED ISSUE #1826: Cache overrides empty string
So if you pass jade an empty string to render, and also pass it a `filename` in the options, rather than returning an empty string as you would expect, it reads the file and returns the file's contents. For example:

```js
console.log(jade.render('', { filename: 'some_file_path' }));
```

This will log out the contents of the file at the path, rather than an empty string. Any content at all in the string will negate this:

```js
console.log(jade.render('p hello', { filename: 'some_file_path' }));
```

This renders the correct string, regardless of whether that's actually the contents of the file, as you'd expect.

TimothyGu:
Fixes #1826.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -168,7 +168,7 @@ function handleTemplateCache (options, str) {
   if (options.cache && exports.cache[key]) {
     return exports.cache[key];
   } else {
-    if (!str) str = fs.readFileSync(options.filename, 'utf8');
+    if (str === undefined) str = fs.readFileSync(options.filename, 'utf8');
     var templ = exports.compile(str, options);
     if (options.cache) exports.cache[key] = templ;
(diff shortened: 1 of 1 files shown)


TimothyGu:
Travis is dumb.

lib/index.js
```diff
@@ -168,7 +168,7 @@ function handleTemplateCache (options, str) {
   if (options.cache && exports.cache[key]) {
     return exports.cache[key];
   } else {
-    if (!str) str = fs.readFileSync(options.filename, 'utf8');
+    if (str === undefined) str = fs.readFileSync(options.filename, 'utf8');
```
hemanth: heh heh serious? Travis :P

Issue #820, opened 2012-11-20
Maintain Path Hierarchy


vanschroeder:
Use case: Nested template folder structure for source

./src
- template1.jade
  /sub-dir
  - template2.jade

Expected render results
./dest-dir
- template1.html
  /sub-dir
  - template2.html

Actual results
./dest-dir
- template1.html
- template2.html

the inability to maintain nested paths is a huge hinderance to projects that are of non-trivial scope and which require a folder structure to help maintain developer sanity.

PR #1829, opened 2015-01-19
Jade CLI to Maintain Path Hierarchy (Another Attempt with Tests)

seanho:
Change jade cli to keep subdirectory structure in output directory.

I've seen several attempts to solve this problem #820, #822, #1525 but none got merged eventually. Here's my attempt to bring this feature to Jade CLI. Tests have been added to make sure everything's ok.

### DIFF
--- a/bin/jade.js
+++ b/bin/jade.js
@@ -129,7 +129,9 @@ if (files.length) {
       process.exit(1);
     });
   } else {
-    files.forEach(renderFile);
+    files.forEach(function (file) {
+      renderFile(file);
+    });
   }
   process.on('exit', function () {
     console.log();
@@ -171,7 +173,7 @@ function stdin() {
  * Always walk the subdirectories.
  */
 
-function renderFile(path) {
+function renderFile(path, rootPath) {
   var re = /\.jade$/;
   var stat = fs.lstatSync(path);
   // Found jade file/\.jade$/
@@ -189,7 +191,13 @@ function renderFile(path) {
     else                     var extname = '.html';
 
     path = path.replace(re, extname);
-    if (program.out) path = join(program.out, basename(path));
+    if (program.out) {
+      if (typeof rootPath !== 'undefined') {
+        path = path.replace(rootPath, program.out);
+      } else {
+        path = join(program.out, basename(path));
+      }
+    }
     var dir = resolve(dirname(path));
     mkdirp.sync(dir, 0755);
     var output = options.client ? fn : fn(options);
@@ -200,7 +208,9 @@ function renderFile(path) {
     var files = fs.readdirSync(path);
     files.map(function(filename) {
       return path + '/' + filename;
-    }).forEach(renderFile);
+    }).forEach(function (file) {
+      renderFile(file, rootPath || path);
+    });
   }
 }
 
--- a/bin/jade.js
+++ b/bin/jade.js
@@ -193,7 +193,7 @@ function renderFile(path, rootPath) {
     path = path.replace(re, extname);
     if (program.out) {
       if (typeof rootPath !== 'undefined') {
-        path = path.replace(rootPath, program.out);
+        path = join(program.out, path.replace(rootPath, ''));
       } else {
         path = join(program.out, basename(path));
       }
--- a/test/command-line.js
+++ b/test/command-line.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var fs = require('fs');
+var mkdirp = require('mkdirp');
 var path = require('path');
 var assert = require('assert');
 var cp = require('child_process');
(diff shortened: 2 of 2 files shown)


bin/jade.js
```diff
@@ -129,7 +129,9 @@ if (files.length) {
       process.exit(1);
     });
   } else {
-    files.forEach(renderFile);
+    files.forEach(function (file) {
+      renderFile(file);
+    });
```
TimothyGu: Why?

test/command-line.js
```diff
@@ -42,13 +42,15 @@ function run(args, stdin, callback) {
   }, callback);
 }
 
-try {
-  fs.mkdirSync(__dirname + '/temp');
-} catch (ex) {
-  if (ex.code !== 'EEXIST') {
-    throw ex;
+['/temp', '/temp/inputs', '/temp/inputs/level-1-1', '/temp/inputs/level-1-2', '/temp/outputs'].forEach(function (path) {
```
TimothyGu: You can use mkdirp to simplify this a little but.

TimothyGu:
In general this looks good; but because this breaks existing behavior it must wait until Jade v2.

bin/jade.js
seanho: Because forEach passes index as second argument, which is unwanted in this case.

TimothyGu: OK

TimothyGu:
LGTM.

ForbesLindesay:
You'll need to make sure program.out is normalised properly to work on Windows as well as *nix.

Other than that, nothing jumps out at me as a problem.

seanho:
program.out is specified by user, wouldn't they put in correct path format just like they would do with cd command?

ForbesLindesay:
On windows they will likely expect to be able to use `/` and `\` interchangeably, and it is generally necessary if they are to write bash scripts to run on both Windows and *nix

test/command-line.js
```diff
@@ -95,6 +92,34 @@ describe('command line with HTML output', function () {
...
+      assert(html === '<div class="foo">bar</div>');
+      done();
+    });
+  });
+  context('when input is directory', function () {
+    it('jade --no-debug --out outputs inputs', function (done) {
+      fs.writeFileSync(__dirname + '/temp/inputs/input.jade', '.foo bar-1');
+      fs.writeFileSync(__dirname + '/temp/inputs/level-1-1/input.jade', '.foo bar-1-1');
+      fs.writeFileSync(__dirname + '/temp/inputs/level-1-2/input.jade', '.foo bar-1-2');
+      fs.writeFileSync(__dirname + '/temp/input.js', 'throw new Error("output not written");');
```
TimothyGu: This is broken.

You need:

```js
      fs.writeFileSync(__dirname + '/temp/outputs/input.html', 'BIG FAT HEN 1');
      fs.writeFileSync(__dirname + '/temp/outputs/level-1-1/input.html', 'BIG FAT HEN 1-1');
      fs.writeFileSync(__dirname + '/temp/outputs/level-1-2/input.html', 'BIG FAT HEN 1-2');
```

```pcre
s/BIG FAT HEN/something more serious/g
```

TimothyGu:
@ForbesLindesay Do you think it would be a good idea to enable this feature using an option? This way we don't have to wait until jade 2 to finally merge this.

ForbesLindesay:
It shouldn't be a breaking change should it?  The current fact that it doesn't maintain path higherachy is a bug?  I definitely don't want to wait till 2.0 for this.

seanho:
Absolutely don't want to wait. Indeed, the current behaviour is undesired if it's not a bug. A very common and valid scenario like below will cause one file overriding the other in the output folder
blog/
---- posts/
-------- new.jade
---- tags/
-------- new.jade

The Grunt and Gulp plugins for Jade do maintain path hierarchy and I assume a lot of people will expect the same with the CLI.

test/command-line.js
seanho: Do you mean adding the above lines as false output?

TimothyGu: @seanho yes, to "poison" the output so that it can be detected if the file did not change.

TimothyGu:
I agree that in some aspects it is a bug, but thing is it does output the files, just to different dirs. Say you have this:

```
views/
|- input1.jade
|+ input-dir/
 | input2.jade
```

It will output as this:

```
views/
|- input1.html
|- input2.html
```

This is not a bug. It's just a different behavior. This patch makes it look like

```
views/
|- input1.html
|+ input-dir/
 | input2.html
```

@ForbesLindesay I understand you don't want to wait until v2, so that's why I suggested to add an option that enables this new behavior.

TimothyGu:
@ForbesLindesay have you gotten a chance to read my comment above? I might make some changes to the CLI testing code and I don't want to deal with any merge conflicts.

TimothyGu:
@seanho Travis complains about a missing `level-1-1/input.html`

ForbesLindesay:
Fair enough, we can have an option for now and mark it as becoming the default in 2.0.0

TimothyGu:
Obsolete in favor of #1889 which adds an option.

Issue #1514, opened 2014-05-05
Inline textarea bug


valtido:
if I do the following: `.row:	textarea(placeholder="Description")= item.description`

this will output raw text `(placeholder="Description") value=item.description`

expected `<textarea ...>...</textarea>`

however if I did 

    .row
          textarea(placeholder="Description")= item.description

It works fine.

Ask away if you should need anymore information 

I think it's just the `textarea` as my `input` are working fine as in 

`.row: input(value=item.title)`

ForbesLindesay:
Closing as #1835 should fix this.

PR #1835, opened 2015-01-21
Add warning for `:` not followed by a space

ForbesLindesay:
Lets disallow `:` without a space after it as part of the move to 2.0.0.  It would catch #1514 and also makes it easier to tell the difference between `:` and `:filter` when reading jade code (and parsing jade code).


hemanth:
:+1:

PR #1836, opened 2015-01-21
Fix error checking for mixin blocks in nested mixins

### LINKED ISSUE #1552: Error checking for anonymous blocks doesn't support nested mixins
```js
//- mixin declaration
mixin imageAndCaption()
    mixin image
      //- this will get overridden
      img(src="/my/placeholder.png")
    mixin caption
      //- this will get overridden
      p Un-captioned image
    block
    div.someClass
        +image()
        div.myCaption
            +caption()

//- use the mixin
+imageAndCaption()
    mixin image
        img(src="/my/image.png")
    mixin caption
        p this is my caption
```

results in "Anonymous blocks are not allowed unless they are part of a mixin."

(see https://github.com/visionmedia/jade/issues/631#issuecomment-45448961)

ForbesLindesay:
[closes #1552]

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -27,7 +27,7 @@ var Parser = exports = module.exports = function Parser(str, filename, options){
   this.mixins = {};
   this.options = options;
   this.contexts = [this];
-  this.inMixin = false;
+  this.inMixin = 0;
   this.dependencies = [];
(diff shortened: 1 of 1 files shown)


hemanth:
@ForbesLindesay is on a roll! :)

Issue #1794, opened 2014-12-23
Duplicate class attributes


EloB:
Made a small test that duplicates class names.

https://gist.github.com/EloB/85f06cc95276223e2486

EloB:
Is this the expected behaviour?

ForbesLindesay:
No, it's a bug in our attribute merging that assumes the value passed to `&attributes` won't be reused (I think)

EloB:
Okay

ForbesLindesay:
This will be fixed in the next version of jade.

EloB:
Thanks! :)

PR #1837, opened 2015-01-21
Bug 1794

ForbesLindesay:
This doesn't tackle the root cause of #1794 but it does add an acceptable temporary fix/workaround.  All this code should be re-written as part of the 2.0.0 release anyway.


hemanth:
:+1:

PR #1839, opened 2015-01-21
Add support for comments in a case block

### LINKED ISSUE #1707: Comments in case are not handled correctly
Comments are not supported in a case block:

```jade
doctype html
html
  body
   - var s = 'this'
   case s
     //- Comment
     when 'this'
       p It's this!
     when 'that'
       p It's that!
```

shows:

```
Jade:6
    4|    - var s = 'this'
    5|    case s
  > 6|      //- Comment
    7|      when 'this'
    8|        p It's this!
    9|      when 'that'

Unexpected token "comment", expected "when", "default" or "newline"
```

ForbesLindesay:
[fixes #1707]

This treats both buffered and unbuffered comments as unbuffered comments.  I don't think that matters though.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -297,6 +297,7 @@ Parser.prototype = {
     this.expect('indent');
     while ('outdent' != this.peek().type) {
       switch (this.peek().type) {
+        case 'comment':
         case 'newline':
           this.advance();
           break;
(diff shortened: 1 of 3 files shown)


hemanth:
`this.advance();` ;)

Issue #1813, opened 2015-01-08
Reverse of #{}


jacobp100:
`#{}` evaluates code when there's non-code entry. How about `${}` to evaluate jade when there's code entry.
 
E.g.
```
script.
   var x = 3;
   var y = '${p.test Test}'
```

jacobp100:
As an alternative to #1812

TimothyGu:
Sorry but I do not understand what the problem is here. What is your expected HTML output?

Doesn't this meet your standards?

```jade
script.
  var x = 3;
  var y = '#{p.test} Test'
```

vendethiel:
I think you're looking for #[]

ForbesLindesay:
```jade
script.
   var x = 3;
   var y = '#[p.test Test]'
```

does exactly what you're after.

jacobp100:
Thanks!

jacobp100:
How would I go about manipulating this? E.g. Escaping quotes as follows:
```jade
var y = "#[p.test Test].replace('\"', '\\"')"
```

TimothyGu:
```jade
var y = "#[p.test Test]".replace('\"', '\\"')
```

should suffice.

ForbesLindesay:
@TimothyGu I don't think that will work as intended.  I think the short answer is that there is no easy syntax for it.  You will have to use single quotes `'` outside the jade (because jade uses double quotes `"` internally) and then be sure to escape any occurrences of single quotes inside the template.  In short, it's not really a supported use case.

TimothyGu:
^^ True.

Issue #1620, opened 2014-08-08
Operators should all be documented


dlowerre:
I am new to Jade, and I find it very frustrating that there appears to be no explanation in the docs for operators that have special meaning to Jade.  

i.e.  I only know the difference between #{} and !{} because I found a tutorial that just happened to mention it.  This should be well documented.

Also: there really should be a section on 'variables' in the docs.

TimothyGu:
Already documented at http://jade-lang.com/reference/interpolation/ since a long time ago.

PR #1843, opened 2015-01-23
Interpolation reference

enlore:
As cool as this feature is, I feel like it needs it's own reference
page.

Aimed at #1620 and #1813.

### DIFF
--- a/docs/views/reference.jade
+++ b/docs/views/reference.jade
@@ -18,6 +18,7 @@ block content
         +link('filters')
         +link('includes')
         +link('inheritance')
+        +link('interpolation')
         +link('iteration')
         +link('mixins')
         +link('plain text')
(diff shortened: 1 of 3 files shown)


TimothyGu:
You can probably remove and/or merge the chapter from https://github.com/jadejs/jade/blob/master/docs/views/reference/code.jade#L88-L128.

enlore:
Ah. I was using the published content on the web as a reference.  I'm going to dig through the source a little harder as I do this.  How do you guys go about publishing changes to the site?

TimothyGu:
@enlore, we usually deploy new documentation with every new release.

TimothyGu:
@enlore thanks!

enlore:
No prob.

ForbesLindesay:
We should also think about adding some "see also" links to relevant places so people find this when they don't necessarily know what "interpolation" means.

PR #1791, opened 2014-12-19
Refactor

ForbesLindesay:
This is no where near ready to merge, I just wanted an easier way to view the diff.

PR #1850, opened 2015-01-28
Grand v2 Refactoring 

TimothyGu:
Migrated from #1791.


ForbesLindesay:
I think what we'll actually want to do is branch v1 pretty soon and then merge this into master (along with anything currently in v2)

ForbesLindesay:
P.S. I want to deal with the merge conflicts anyway as some of them are new test cases, and most of the others should be easy to resolve.

bin/jade.js
```diff
@@ -58,6 +59,17 @@ program.on('--help', function(){
 
 program.parse(process.argv);
 
+// locals given, parse them
+
+var locals;
+if (program.locals) {
+  if (exists(program.locals)) {
+    locals = JSON.parse(fs.readFileSync(program.locals));
```
alejonext: Well, if you're going to get is better to use the `require` ?

ForbesLindesay: I wouldn't bother reviewing the changes in this file too much.  There are lots of changes in master that will need merging with a few changes in this, so what's here at the moment isn't particularly representative of what I expect the finished product to look like.

alejonext: OK! It is not to bother ... I do it in the best of the spirits! is that your work is hard!

ForbesLindesay: Didn't mean that to come across as a complaint.  It's always good to get feedback, just thought if you were reviewing the code I didn't want you to waste your time :)

TimothyGu:
#1972

Issue #1086, opened 2013-06-19
watch father file of includes


acidjazz:
I have to use jade command-line mostly because of other languages I have to use for my current line of work (same with stylus).  One feature I would love is for if I'm watching say "index.jade" is for jade to re-compile upon the write of any included file say "_header.jade" or even extension related to index.jade. I noticed stylus does this and I'd love to not have to re-compile jade every time I modify an include.

Thanks!

jayliu50:
@ForbesLindesay could you provide an example of accessing the list of dependencies, please?

Also, are you suggesting that jade watch doesn't recompile extended and included files, by design? For example, I have `page.jade` which extends `layout.jade`. When I made a change to `layout.jade`, `layout.html` is updated, but not `page.html`.

TimothyGu:
@jayliu50 In the latest version of [jade-cli](https://www.npmjs.org/package/jade-cli) everything works as expected, i.e. while `jade -w page.jade` is running, if `layout.jade` is changed `page.html` is updated. OTOH, when you do `jade -w page.jade layout.jade`, when `layout.jade` is changed both HTMLs are updated.

jayliu50:
@TimothyGu Thanks for the response. Yes, it does work. I found out that jade was resolving to an old copy of 1.7 for some reason.

PR #1851, opened 2015-01-28
Fixes for watch mode.

pavel:
Problems that are fixed in the request:

1. #1086
2. problem, that when watching a folder jade-cli was not watching subtrees.

Solutions provided:

1. fetch dependencies from ```jade.compile```, and watch'em.
2. watch every file that is a target for render.

Improvement:

1. Unwatching a file, if the file was moved or removed.

### DIFF
--- a/bin/jade.js
+++ b/bin/jade.js
@@ -102,45 +102,59 @@ options.doctype = program.doctype;
 
 var files = program.args;
 
+var watchList = [];
+
 // compile files
 
 if (files.length) {
   console.log();
   if (options.watch) {
-    files.forEach(function(filename) {
-      function errorToString(e) {
-        return e.stack || /* istanbul ignore next */ (e.message || e);
-      }
-      try {
-        renderFile(filename);
-      } catch (e) {
-        // keep watching when error occured.
-        console.error(errorToString(e));
-      }
-      fs.watchFile(filename, {persistent: true, interval: 200},
-                   function (curr, prev) {
-        // istanbul ignore if
-        if (curr.mtime.getTime() === prev.mtime.getTime()) return;
-        try {
-          renderFile(filename);
-        } catch (e) {
-          // keep watching when error occured.
-          console.error(errorToString(e));
-        }
-      });
-    });
     process.on('SIGINT', function() {
       process.exit(1);
     });
-  } else {
-    files.forEach(renderFile);
+    process.on("exit", function () {
+      watchList.forEach(fs.unwatchFile);
+    });
   }
+  files.forEach(renderFile);
   process.on('exit', function () {
     console.log();
   });
 // stdio
 } else {
   stdin();
+}
+
+function watchFile(path, base) {
+  if (watchList.indexOf(path) !== -1) return;
+  console.log("  \033[90mwatching \033[36m%s\033[0m", path);
+  fs.watchFile(path, {persistent: true, interval: 200},
+               function (curr, prev) {
+    // istanbul ignore if
+    if (curr.mtime.getTime() === prev.mtime.getTime()) return;
+    if (!fs.existsSync(path)) {
+      // file was removed or moved
+      console.log("  \033[90munwatching \033[36m%s\033[0m", path);
+      watchList.splice(watchList.indexOf(path), 1);
+      fs.unwatchFile(path);
+      return;
+    }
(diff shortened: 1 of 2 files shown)


ForbesLindesay:
Travis is failing :(

pavel:
But why???
1257.1 Some timeout issues, when running in non-watch mode.
1257.2 Timeout when running "jade --help" ... how did even I could affect that piece???

pavel:
I ran ```npm run coveralls``` and it succeded. Well partially, int the end it failed, because it was not able to write report or smthn.

bin/jade.js
```diff
@@ -102,39 +103,24 @@ options.doctype = program.doctype;
...
-      });
-    });
     process.on('SIGINT', function() {
       process.exit(1);
     });
-  } else {
-    files.forEach(renderFile);
+    // unwatch what's watched on exit
+    process.on("exit", function () {
+      watchList.forEach(fs.unwatchFile);
```
ForbesLindesay: I don't think we need to do this.  When node exits it will just stop watching the files at that point.

bin/jade.js
```diff
@@ -144,6 +130,51 @@ if (files.length) {
 }
 
 /**
+ * Watch for changes on path
+ * Renders base if specified, otherwise renders path
+ * If watched path was moved or removed, unwatches the path
+ */
+function watchFile(path, base) {
+  path = normalize(path);
```
ForbesLindesay: We should use `resolve` rather than `normalize` here.  Probably doesn't make any difference, but what we want is the full absolute path.

bin/jade.js
```diff
@@ -102,39 +103,24 @@ options.doctype = program.doctype;
...
       process.exit(1);
     });
-  } else {
-    files.forEach(renderFile);
+    // unwatch what's watched on exit
+    process.on("exit", function () {
+      watchList.forEach(fs.unwatchFile);
+    });
   }
+  files.forEach(renderFile);
```
ForbesLindesay: This needs to be `tryRender` in watch mode.

ForbesLindesay:
I've added a few comments.  Other than those, it looks pretty good.

To fix the timeouts, I would suggest going through https://github.com/jadejs/jade/blob/master/test/command-line.js and everywhere we have something like:

```js
  if (isIstanbul) {
    this.timeout(8000);
    this.slow(6000);
  } else {
    this.slow(250);
  }
```

make it something like

```js
  if (isIstanbul) {
    this.timeout(60000);
    this.slow(6000);
  } else {
    this.timeout(30000);
    this.slow(3000);
  }
```

We want timeouts that are way longer than we expect them to take, rather than having them on the brink of failing every time.  Sorry for assuming your change caused this initially.

bin/jade.js
pavel: Let me verify it with the node's source code, first.

bin/jade.js
pavel: ```path``` variable is getting console logged. If you're ok with seeing absolute pathes on the log, then I'll change that.

pavel:
Sure, I'll increase the timeouts in this pull-request. Don't be sorry, you do what you gotta do.

bin/jade.js
ForbesLindesay: Personally I don't really care what we log, @TimothyGu any thoughts? I guess normalize should still work.

pavel:
I've left ```normalize```, if anyone thinks that it's important to switch to ```resolve```, then let me know.

bin/jade.js
TimothyGu: I don't care either, but `normalize` is probably a little bit better for readability and compat (not like anyone would care).

TimothyGu:
@pavel I don't really like the test timeouts. 30 seconds is *way* too much for the timeouts, and 3 seconds for being slow isn't that reasonable either.

pavel:
I did what @ForbesLindesay suggested.
@TimothyGu  What are your suggestions?

TimothyGu:
@pavel how about this:

  | istanbul | non-istanbul
--|------------|------------------
TO | 20s | 12.5s
slow | 3s | 2s

pavel:
Sure, let me update.

TimothyGu:
Cool. Merged, thanks!

Issue #1854, opened 2015-01-30
Watcher unwatches after saving jade file with VIM


pavel:
The thing I've discovered after my pull-request #1851 was merged.
This affects only VIM users, and basically because of the awkwardness of VIM's save.

I'll fix it and create pull request.
Sorry for the inconvenience.

TimothyGu:
Thanks for spotting this. I had the same experience when I was doing abff32cd5f5f02df4724031c139003cf56a42c96.

pavel:
Also when saving with VIM, sometimes jade throws ENOENT error for a sec, and then another watch event arrives, and the file renders fine. I'll fix it as well, to  prevent this awkward error from being logged.

TimothyGu:
Basically it seemed that vim deletes the file, resulting in an ENOENT and/or `fs.existSync()` failure.

pavel:
@TimothyGu Right, that's what happening.

PR #1855, opened 2015-01-30
Fix for #1854

pavel:
Solution:
Wait for 200ms, and then check if the file exists.
And when the file is gone, unwatch it.

### DIFF
--- a/bin/jade.js
+++ b/bin/jade.js
@@ -126,6 +126,15 @@ if (files.length) {
 }
 
 /**
+ * Unwatches a file, specified by a path
+ */
+function unwatchFile(path) {
+  console.log("  \033[90munwatching \033[36m%s\033[0m", path);
+  watchList.splice(watchList.indexOf(path), 1);
+  fs.unwatchFile(path);
+}
+
+/**
  * Watch for changes on path
  * Renders base if specified, otherwise renders path
  * If watched path was moved or removed, unwatches the path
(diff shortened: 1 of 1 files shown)


TimothyGu:
LGTM. @ForbesLindesay any comments?

ForbesLindesay:
Can't we just keep watching it forever?  That 200ms timeout is a very arbitrary time, and I don't really see the problem with just continuing to watch it indefinitely.  It wastes a tiny bit of resources, but only in development and you can always restart the CLI if it becomes a serious problem.

pavel:
Yep, that sounds reasonable. I'll fix and update.

pavel:
@ForbesLindesay @TimothyGu 
Do we need to suppress ENOENT errors? They happen when a watched file has gone, that pops frequently when editing with VIM.

TimothyGu:
@pavel, @ForbesLindesay meant to just remove the unwatching routine. It's kind of hard to fix.

ForbesLindesay:
We probably should wrap the fs.readFileSync in a try catch and ignore ENOENT errors but only when it's as a result of a watch triggering.

TimothyGu:
Thanks!

ForbesLindesay:
Can we add a test case for this? Then remove the additional Istanbul annotation?

TimothyGu:
@ForbesLindesay Already removed 1 minute before your comment: db894a1c5df0b85302c9a7ce66c67037ee347c72

I'll work on adding a test.

PR #1861, opened 2015-02-07
Travis: test with io.js 1.1 instead of Node.js 0.11

TimothyGu:
Node.js 0.11 is dead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: node_js
 node_js:
   - "0.10"
-  - "0.11"
+  - "0.12"
+  - "iojs-1.1"
 
 # Use faster Docker architecture on Travis.
 sudo: false


hemanth:
Long Live...

TimothyGu:
Actually Node.js 0.12 [just got released today](http://blog.nodejs.org/2015/02/06/node-v0-12-0-stable/).

rlidwka:
LGTM

Maybe leave 0.11 as is, but we need to test for io.js anyway.

TimothyGu:
@rlidwka Actually I added a 0.12 target. Nobody should be using 0.11.

jonathanong:
gotta test 'em all!

TimothyGu:
@jonathanong At least this is better than those humongous sauce labs "badge."

PR #1869, opened 2015-02-13
Add client-compilation example

TimothyGu:
See #1867.

### DIFF
--- a/examples/client-compilation.jade
+++ b/examples/client-compilation.jade
@@ -1,0 +1,140 @@
+//
+  This example shows how to compile and render Jade templates in the browser.
+
+  Note that although it is supported, we do not *recommend* compiling in the
+  browser; rather, you should pre-compile the template using compileClient
+  or compileFileClient.
+
+//-
+  You can either compile this to examples/client-compilation.html and open it
+  manually with a web browser, or you can serve this file using the server
+  examples/client-compilation.js.
+
+doctype html
+html
+  head
+    title Jade compilation demo
+    style.
+      * {
+        box-sizing: border-box;
+      }
+      iframe, pre, textarea {
+        height: auto;
+        border: 1px solid #a9a9a9;
+        width: 100%;
+        max-height: 600px;
+        min-height: 120px;
+      }
+      /* Preformatted <textarea> */
+      textarea.pre {
+        font-family: monospace;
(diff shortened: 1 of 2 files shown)


examples/client-compilation.jade
```diff
@@ -0,0 +1,140 @@
+//
+  This example shows how to compile and render Jade templates in the browser.
+
+  Note that although it is supported, we do not *recommend* compiling in the
+  browser; rather, you should pre-compile the template using compileClient
+  or compileFileClient.
```
ForbesLindesay: It might be better to actually put this text as a paragraph on the HTML page as so people see it if they open the sample in their browser.

examples/client-compilation.jade
```diff
@@ -0,0 +1,140 @@
...
+        "name": "Timothy"
+      }
+
+    h2 Raw HTML
+    pre#output
+
+    h2 Rendered HTML
+    iframe#output-html
+
+    script(src='../jade.js')
```
ForbesLindesay: This doesn't look like it is the right path if you're serving the file via the express web server.

examples/client-compilation.jade
```diff
@@ -0,0 +1,140 @@
...
+    script.
+      function comp() {
```
ForbesLindesay: Since this script is at the bottom of the page, we shouldn't need to wait for onload

ForbesLindesay:
This looks good, other than those small points :)

examples/client-compilation.jade
TimothyGu: Sounds good

examples/client-compilation.jade
TimothyGu: Yes, it is not the correct path. I did this because of two reasons:

1. it will work if the user render the template to examples/client-compilation.html
2. it will work even in a server environment because the page is served as `http://127.0.0.1/`, and its `../jade.js` is rendered as `http://127.0.0.1/jade.js` (because there is no higher level). I agree this is a hack however.

examples/client-compilation.jade
TimothyGu: Ah yes true. I didn't realize one could just `comp()` at the end of the page :)

TimothyGu:
All comments fixed, merged.

PR #1886, opened 2015-03-08
Add parser support for inline filters like #[:link home]

evanw:
I found myself wanting support for filters inside interpolation nodes. I'm not sure if this feature was intentionally omitted but if not it seems easy to add. I found this feature useful for generating and verifying internal links among templates, sort of like links inside a wiki. If this pull request isn't useful then feel free to close it.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -684,6 +684,12 @@ Parser.prototype = {
     var block = new nodes.Block;
     block.line = this.line();
     var body = this.peek();
+    if (body.type === 'text') {
+      var text = new nodes.Text(this.expect('text').val);
+      text.line = block.line;
+      block.push(text);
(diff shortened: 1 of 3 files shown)


TimothyGu:
Doesn't look wrong to me, but I am not the one to review code relating to parser stuff.

CC @ForbesLindesay.

ForbesLindesay:
I definitely like the change.  My concern is that I don't really understand how it works :s (my failing not yours).  I wonder if you could add the same tests/fixes to https://github.com/jadejs/jade-lexer and https://github.com/jadejs/jade-parser

TimothyGu:
Now that jadejs/jade-parser#5 is merged, this is obsolete.

PR #1899, opened 2015-03-23
Grammar fixes.

davezuko:
Use possessive form of "its" rather than "it's". Removed an unnecessary (read: ungrammatical) comma from same paragraph.


TimothyGu:
Thanks!

PR #1917, opened 2015-04-01
no value attributes are rendered differently

leider:
depending on wether it is done via included mixin or directly.

see `<div ui-view></div>` in the example

### DIFF
--- a/test/cases/attr-no-value-mixin.html
+++ b/test/cases/attr-no-value-mixin.html
@@ -1,0 +1,7 @@
+<!DOCTYPE html>
+<html>
+  <head></head>
+  <body>
+    <div ui-view></div>
+  </body>
+</html>
--- a/test/cases/attr-no-value-mixin.jade
+++ b/test/cases/attr-no-value-mixin.jade
@@ -1,0 +1,4 @@
+include auxiliary/mixin-attr-no-value
+doctype html
+html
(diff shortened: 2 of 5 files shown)


TimothyGu:
This has nothing to do with if a mixin is used.

For example, the following works:

```jade
doctype html

mixin attr-mixin
  head
  body
    div(ui-view)
html
  +attr-mixin()
```

The reason why this fails is that the mixin is in a different file not declared with `doctype html`.

leider:
So there currently is no way of putting the fragment in its own file?

That is "sad". But it is not really a showstopper. For me it's ok if you close the PR and the corresponding PR.

leider:
Now that is weird. This also works:

```jade
doctype html
include auxiliary/mixin-attr-no-value
html
  +attr-mixin()
```

Just moving the include below the doctype.

TimothyGu:
I am going to close this, since #1918 is still opened. Also, you can use the `doctype` option to work around the bug:

```jade
include auxiliary/mixin-attr-no-value
doctype html
html
  +attr-mixin()
```

```sh
$ jade --doctype html attr-no-value-mixin.jade
```

```js
var jade = require('jade')
var fs = require('fs')

jade.renderFile('attr-no-value-mixin.jade', { doctype: 'html' })
```

PR #1920, opened 2015-04-02
HACK: Pin to earlier acorn-globals and with version

TimothyGu:
Works around marijnh/acorn#232 and make Travis useful again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,13 +22,14 @@
     "jade": "./bin/jade.js"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
I think for now we should just be patient as acorn is very active so it shouldn't take long to get the fix merged.

ForbesLindesay:
These have been updated now, all is well :)

TimothyGu:
Thank you :)

PR #1928, opened 2015-04-12
Add support for jstransformers, deprecate Transformers

ForbesLindesay:
So many people have done such fantastic work to make this possible.  This will make it really easy to independently publish/update/fix each separate filter.


lib/filters.js
```diff
@@ -1,14 +1,96 @@
...
+  coffeekup: 'coffeecup',
+  // The `style` transformer is not the same as the `stylus` jstransformer
+  styl: 'stylus',
+  coffee: 'coffee-script',
+  coffeescript: 'coffee-script',
+  coffeeScript: 'coffee-script',
+  // these marker transformers haven't made sense in a long time
+  css: 'verbatim',
+  js: 'verbatim',
+};
```
RobLoach: Might be able to use the `inputFormats` values for these `alternatives`.

test/cases/filters.coffeescript.jade
```diff
@@ -1,6 +1,6 @@
 script(type='text/javascript')
-  :coffeescript
+  :coffee-script
     regexp = /\n/
-  :coffeescript(minify=true)
+  :coffee-script(minify=true)
```
RobLoach: Added some additional `inputFormats` to the coffeescript transform:
https://github.com/jstransformers/jstransformer-coffee-script/commit/d954505f2e3e99b53b99cb3479829b9ede04e95a

lib/filters.js
ForbesLindesay: We could, but I think in this case it's probably better to get people to specify the exact canonical name of that transform.  It prevents any potential ambiguity and makes loading the relevant transformer much simpler (otherwise we have to look for all possible transformers in the parent path and load all of them just in case one of them matches).

Issue #1931, opened 2015-04-14
Whitespace in textarea


shrenikp:
[This commit](https://github.com/jadejs/jade/commit/0a78e7ff81253b9b2bd8b2f699ceb98d07882aeb) doesn't seem to be merged. Still having the whitespace issue in `textarea` as per #500

ivanjuras:
+1, doesn't seem to be working for me either.

ForbesLindesay:
Pull request welcome.

TimothyGu:
#1935

groenroos:
+1. @ForbesLindesay

ForbesLindesay:
`'textarea'` should be marked as whitespace sensitive in the same way that `'pre'` is in https://github.com/jadejs/jade-code-gen/blob/master/index.js

PR #1935, opened 2015-04-17
Mark <textarea> as an inline tag

### LINKED ISSUE #500: pretty + textarea = rogue spaces!
I'm using pretty printed html in my dev environment and I'm getting textarea elements with the close dropped to the next line and matching the indentation, which consequently results in a bunch of spaces that shouldn't be there.

TimothyGu:
Closes #500 and #1931.

### DIFF
--- a/lib/inline-tags.js
+++ b/lib/inline-tags.js
@@ -20,4 +20,5 @@ module.exports = [
   , 'strong'
   , 'sub'
   , 'sup'
+  , 'textarea'
 ];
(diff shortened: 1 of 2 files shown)


TimothyGu:
Sorry about the test failure. It for some reason passed here. Fixed.

ForbesLindesay:
Shouldn't this be handled in the same way as `'pre'` is in https://github.com/jadejs/jade/blob/master/lib/compiler.js ?  `<textarea>` isn't really an inline tag, it just has white space sensitive content.

PR #1952, opened 2015-05-16
disable compileDebug for express' production env

alubbe:
Building on the discussion in https://github.com/jadejs/jade/pull/1951, I would recommend to disable compileDebug by default for express' production env.

That's really all there is to it - what do you think?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -410,4 +410,9 @@ exports.compileFileClient = function(path, options){
  * Express support.
  */
 
-exports.__express = exports.renderFile;
+exports.__express = function(path, options, fn) {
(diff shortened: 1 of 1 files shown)


alubbe:
The failing test has nothing to do with the code here, it has been fixed at https://github.com/jadejs/jade/pull/1953

ForbesLindesay:
I have considered this in the past, but didn't get round to doing it because I did some initial experiments and found a pretty negligible performance difference.  However, with the recent memory management issues, it's clearly become more of a problem.  I can't really see a good reason not to do this.  I also think the approach of adding it only when jade is used implicitly (via `__express`) is probably a good idea.

PR #1958, opened 2015-05-20
Allow % symbol in class names.

naturalethic:
The skel css framework (https://github.com/n33/skel) uses % symbol in class names.  Unless there is a good reason to prevent using the % symbol in class names in jade, I recommend this be merged.

### DIFF
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -271,7 +271,7 @@ Lexer.prototype = {
    */
 
   className: function() {
-    return this.scan(/^\.([\w-]+)/, 'class');
+    return this.scan(/^\.([\w-%]+)/, 'class');
   },
 
   /**


naturalethic:
Well, it occurs to me that the skel framework also requires the ability to put parens ( ) in class names.  Given that parens open the attributes in jade, I'm not sure this one is even possible.  If you can't make jade fully compatible with skel, perhaps there is no other reason to provide the % either.  It seems that using skel precludes using jade.  Any ideas?

naturalethic:
I've opened a ticket in the skel repo to discuss this issue there:

https://github.com/n33/skel/issues/136

alubbe:
Yay for escaped character names in class. (see more @ https://mathiasbynens.be/demo/crazy-class)
One way to do all of this crazyness right now is to use ```p(class="pe%ter")``` instead of the ```.``` notation. I'm not sure we want to encourage this by including specific characters. I think that it would make more sense to think about arbitrary escapes the way that CSS itself does it, so that this would be jade's way:
```
p.pe\%ter lorem ipsum
```
See http://codepen.io/anon/pen/dopOar

Finally, a quick question. Could you point me to where skel actually uses these class names? On their website I only found 'normal' class names  http://getskel.com/assets/css/style.css

jeromew:
Since the `p(class="pe%ter")` notation is a way to accept skel notation (including parentheses), I vote for not merging this PR at this stage (closing it and maybe re-open in at a future stage). Are there other css libraries that use such meta-naming ?

the 'standard' valid characters for a class name are detailed on http://stackoverflow.com/questions/448981/what-characters-are-valid-in-css-class-names-selectors

+1 on accepting valid CSS class names (if escapes are not already parsed)

ForbesLindesay:
@jeromew I know some css libraries use `/` for "namespacing" css properties.

I'm not sure adding support for more complex class names makes too much sense as I'm concerned about the readability (humans must be able to parse the code and see where one class name begins and another one ends).  I therefore don't want to support any punctuation (other than what's already permitted) at this time.  I do, however, want to support letters that have accents, and possibly other things like ß (providing that's valid in css class names).

PR #1959, opened 2015-05-21
1.10.0

alubbe:
See https://github.com/jadejs/jade/issues/1956

I'm assuming the updated dependencies & the addition of jstransformers since the last release caused the big rewrite of the compiled jade.js and the reworked first line of runtime.js.

Please help me in green lighting this, so we can this show on the road.
Next, I would look at the docs to update the new behaviour. I would add a comment about compileDebug's standard behaviour in express, can I interest someone else in documenting the new transformers and the cli hierarchy flag? :)


alubbe:
1.10.0 is now released - any takers for updating docs for the new transformers options or the cli hierarchy flag?

PR #1961, opened 2015-05-23
Fix comma splice in inheritance doc

vergenzt:
These sorts of things just always bug me. :)


TimothyGu:
> These sorts of things just always bug me.

Same here :)

Issue #796, opened 2012-10-31
Allow multiline JS var definition


kinergy:
Similar to how this works in Jade:

    input(
      type='checkbox'
      name='agreement'
      checked)

I'd like to be able to do this:

    config = {
      a: 'something',
      b: 'something-else'
    }

instead of having to write:

    config = { a: 'something', b: 'something-else' }

This is a short example, however as config grows it would be much prettier if funky spacing was allowed.

TimothyGu:
I am going to go ahead and unlock this issue. Please refrain from adding more +1’s, and limit your comments to constructive implementation details.

iliakan:
+1, it worked for me as Forbes said here https://github.com/jadejs/jade/issues/796#issuecomment-39235865, but only with debug disabled.

TimothyGu:
@iliakan it is expected that !debug works as debugging "symbols" are attached in between lines.

iliakan:
@TimothyGu yes, I felt it's the reason :/
Is it possible to do something with it? Maybe something like `js:`, to let jade know we're including a non-breakable js...

TimothyGu:
@iliakan I'm not extremely motivated to do this. We will accept pull requests for the following syntax though:

```jade
-
  var af = {
    prop: 1
  }
```

See https://github.com/jadejs/jade/issues/796#issuecomment-39235865

acidjazz:
If we do need this or think it's helpful I'd say a more generic, faster, and safer implementation would be that of a multi-line JS block similar identifier like /* */ or .  

This sort of also bumps into the factor of functionality balance though, if I have am doing THIS much in jade that I require such a feature, should this be in Jade?

TimothyGu:
@acidjazz said:

> If we do need this or think it's helpful I'd say a more generic, faster, and safer implementation would be that of a multi-line JS block similar identifier like /* */ or .

I'd rather use the syntax proposed by @ForbesLindesay to be more consistent with the currently used `-`.

> This sort of also bumps into the factor of functionality balance though, if I have am doing THIS much in jade that I require such a feature, should this be in Jade?

+1, although in many cases it's just a simply look-up table or loop array, so I guess it's fine.

boompig:
+1

dogancelik:
:+1: I would be happy if someone could implement multi-line JS.

AbraaoAlves:
+1

ironmancris:
+1

alephyud:
[...]

PR #1965, opened 2015-05-27
Added block code support

alephyud:
Added support for multi-line JS code as discussed in [pull request #796.](https://github.com/jadejs/jade/issues/796#issuecomment-39235865)

Block code starts with a line containing only '-', possibly with trailing whitespace. A block after that line is treated as unbuffered Javascript code, such as:

    -
      var arr = ["Very", "Long",
                 "Array"];

Changed tests and docs accordingly.

Please let me know if there are any issues - thank you.

### DIFF
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -586,6 +586,21 @@ Lexer.prototype = {
     }
   },
 
+
+  /**
+   * Block code.
+   */
+
+  blockCode: function() {
+    var captures;
+    if (captures = /^-[ \t]*(\n|$)/.exec(this.input)) {
+      this.consume(captures[0].length - captures[1].length);
+      var tok = this.tok('blockCode');
+      this.pipeless = true;
+      return tok;
+    }
+  },
+
   /**
    * Attributes.
    */
@@ -916,6 +931,7 @@ Lexer.prototype = {
       || this["while"]()
       || this.tag()
       || this.filter()
+      || this.blockCode()
       || this.code()
       || this.id()
       || this.className()
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -593,8 +593,8 @@ Lexer.prototype = {
 
   blockCode: function() {
     var captures;
-    if (captures = /^-[ \t]*(\n|$)/.exec(this.input)) {
-      this.consume(captures[0].length - captures[1].length);
+    if (captures = /^-\n/.exec(this.input)) {
+      this.consume(captures[0].length - 1);
       var tok = this.tok('blockCode');
       this.pipeless = true;
       return tok;
--- a/docs/views/reference/code.jade
+++ b/docs/views/reference/code.jade
@@ -21,6 +21,27 @@ block documentation
           <li>item</li>
           <li>item</li>
           <li>item</li>
+  
+  p Jade also supports block unbuffered code:
+  
+  .row(data-control='interactive')
+    .col-lg-6
+      +jade
+        :jadesrc
+          -
+            list = ["Uno", "Dos", "Tres",
+                    "Cuatro", "Cinco", "Seis"]
+          each item in list
+            li item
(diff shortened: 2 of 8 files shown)


alephyud:
Note: Travis build on Node 0.10.0 failed with error message `Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}`. As I understand, such responses may sometimes spontaneously arise due to issues of Travis / Coveralls integration, could someone please re-run the checks? Many thanks.

ForbesLindesay:
That's only coveralls, so I'm not too worried about that error.  The actual tests passed.

ForbesLindesay:
Thanks so much for this, it's absolutely awesome work.  A couple of small thoughts though:

1. We don't normally like /jade.js to be re-built as part of pull requests, but I can always revert that myself when I merge so you don't need to worry about it.
2. If we changed `/^-[ \t]*(\n|$)/` to `/^-\n/` it would be a non-breaking change, since that would currently be a syntax error.  (i.e. disallow white space after the `-`).  We could then warn about `-` followed by white space and ultimately make it an error in a future version.  Can you make this change?

Would you like to be added as a collaborator on this project, to help review and merge pull requests?

alephyud:
Thank you! Understood about /jade.js .

On 2) - could you please clarify your idea? Not sure if we really should make trailing whitespace an error: after all, it often appears by accident and makes no harm, and seconly, in the current version strings like `"-    \n"` do not cause a syntax error and are accepted as (empty) code, since any whitespace is parsed as code due to greedy evaluation of `/^(!?=|-)[ \t]*([^\n]+)/` in `Lexer.prototype.code()`. So, with `/^-\n/` pattern for block code, a code like

    - [whitespace]
      p=5

would yield different results from 

    -
      p=5

not something we would probably want...

ForbesLindesay:
We can't make `- [whitespace]` an error yet, because that would be a breaking change.  Unfortunately, because it is currently already valid, we also can't change its behaviour.

In the short term, we could make `-[no whitespace]` work the way you want it to, because that is currently a syntax error.  In the long term, we could make `- [whitespace]` do whatever we want, providing we add a warning now.

lib/lexer.js
```diff
@@ -586,6 +586,21 @@ Lexer.prototype = {
...
   },
 
+
+  /**
+   * Block code.
+   */
+
+  blockCode: function() {
+    var captures;
+    if (captures = /^-[ \t]*(\n|$)/.exec(this.input)) {
```
TimothyGu: Do we support using a tab character after the dash? (Like `-\t`)

alephyud: As of now `-\t` will be parsed as source code - so if we want to avoid breaking changes it would probably best to stick to `/^-\n/` as proposed by @ForbesLindesay.

alephyud:
Thank you, understood - sent another commit.

On a related topic, I've noticed that pipeless-text (block code and block comments) is not included into the line count when reporting errors.

It seems that it can be fixed by tweaking `Lexer.prototype.pipelessText` code:
(lines 858-864 in lib/lexer.js)

        var str = this.input.substr(1, i);
        isMatch = str.substr(0, indent.length) === indent || !str.trim();
        if (isMatch) {
          // consume test along with `\n` prefix if match
          this.consume(str.length + 1);
    +     ++this.lineno;
          tokens.push(str.substr(indent.length));
        }

Would take time to test it a little - if it works, shall I include it in this pull request or open a separate one?

TimothyGu:
> Would take time to test it a little - if it works, shall I include it in this pull request or open a separate one?

A separate one please.

lib/lexer.js
TimothyGu: Okay never mind. I misunderstood your discussions above.

CREEATION:
So waiting for this to happen :heart: :hear_no_evil:

Issue #1888, opened 2015-03-11
Rerendering dependencies does not work when child files are already watched


sodoku:
https://github.com/jadejs/jade/issues/1086 addressed watching father files and their children jade files. It does not work for me, if the children are already being watched. Please see the example for more explanations.

# Example

A `main.jade` which includes `a.jade` and `b.jade`

    ├── main.jade
    ├── includes/
        ├── a.jade (included from main.jade)
        ├── b.jade (included from main.jade)

## working

If I watch only the `main.jade`

```bash
jade -w *.jade
```

the files `a.jade` and `b.jade` are automatically watched (as they are a dependency from `main.jade`). If I change `a.jade`, `main.jade` will be rerendered.

## not working

If I watch all files

```bash
jade -w .
```

all files are being watched. But if I now change `a.jade`,  only `a.jade` will be rerendered. `main.jade` will  _not_ be rerendered.

netlemur:
+1

TimothyGu:
Darn it, now we actually have to build a tree of dependency.

distilledhype:
I have the same issue. Is there a current workaround?

TimothyGu:
@distilledhype not yet, until we merge my PR which will be in the next release (hopefully 2.0.0).

distilledhype:
@TimothyGu OK. Good to know that it is fixed already though. Any idea when that will be?

distilledhype:
@ForbesLindesay awesome that this is fixed now, but I assume I will have to wait until the release of Jade 2 in order to use it correct?

ForbesLindesay:
You could try using the new https://github.com/jades/jade-cli which will replace the built in jade binary in 2.0.0.  It should already have all the updates.

PR #1967, opened 2015-05-31
Re-render all base files for a dependency

richardgoater:
Keep an array of base paths per dependency.
Loop over and render each base when dependency changes.

A simple fix for a problem I was having with the watcher - let me know if I need to do anything else or if I'm completely mistaken :)

### DIFF
--- a/bin/jade.js
+++ b/bin/jade.js
@@ -120,6 +120,8 @@ var files = program.args;
 
 var watchList = [];
 
+var watchDependencies = {};
+
 // function for rendering
 var render = program.watch ? tryRender : renderFile;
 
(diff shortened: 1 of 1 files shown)


TimothyGu:
Thanks for doing this. This is #1888, and I remember that I did have a patch for that bug. But now seems like I can't find that branch any more…

This pull request is a little bit more convoluted than it needs to be. For example, `watchList` and `watchDependencies` can just be one object. I'll make a new patch and let's see if my approach is a little bit easier to read :)

TimothyGu:
Okay, new PR with tests at #1968. I believe that one is a lot simpler, but feel free to comment there.

I'll keep this PR open for now.

richardgoater:
Hi @TimothyGu, yes that looks much better! Nice tests also, thanks very much for doing this. I'd really appreciate it if you could credit me for the initial implementation in the changelog please, and feel free to close this PR.

TimothyGu:
Totally. I've updated the PR over there, so I'm going to close this one now.

PR #1971, opened 2015-06-01
travis: Use after_success for coverage report

TimothyGu:
This way the Travis build outcome is not affected by Coveralls server downtime, which unfortunately is an issue.


alubbe:
lgtm
also, yay for needing this...

PR #1972, opened 2015-06-01
Refactor

ForbesLindesay:
Move lexer/parser/compiler into separate modules.

This is very nearly ready to merge.


alubbe:
oh. my.
Pretty crazy - I hope to find time this week to take a look.

ForbesLindesay:
I'm going to try and make this a bit easier to review.  Once it's ready, it can become the new 2.0.0 version of jade.

alubbe:
Without having looked at it in detail - are you also (planning on) touching the way the final compiled function works?
I've started benchmarking the living crap out of jade's functions at https://github.com/alubbe/jade-ejs-bench, http://jsperf.com/jade-better-escape. http://jsperf.com/html-compilers/2 and http://jsperf.com/html-compilers/3
There are two easy performance improvements at https://github.com/alubbe/jade/commit/55a367c3cd05ef480bf15114adbc4b56d24d2ae3 and https://github.com/alubbe/jade/commit/ede9a0740f574af7b8b4dfda7830f439ed5010df that I would PR later.
I just wanted to update you so that we don't end up working on the same stuff ;)

ForbesLindesay:
I'm now finished (at least for now).  As you can see, I've split out everything into separate modules (including the code-gen stage and the jade-runtime).  The second commit has the stuff that really needs reviewing, the first commit is just deleting everything that's no longer used.  We can move more of the code in parse, compile and compileClientWithDependenciesTracked (that currently include some duplicated code) into the jade-code-gen module, which would make it easier to experiment with things like a `jade-optimised-code-gen` module for experimental improvements.

I would be pretty happy to deprecate the support for objects in each.  One way to do that might be to add new syntax like:

```jade
each value of array
```

to match the ES6 syntax.  That way the old syntax could continue to be slow and support objects as well as arrays, while the new syntax could be faster. 

I think there's room for a lot of interesting things around mixins.  I would very much like to see some form of automated in-lining process for mixins that are only being given or mostly being given constants anyway.  We could also investigate something like `jade_mixin_name` in stead of `jade_mixins['name']` as that may be a lot faster.

Anyway, my goal with this pull request was to move towards making it easier to hack on semi-forks of jade without having to necessarily fork the entire project.

lib/index.js
```diff
@@ -196,35 +168,27 @@ function handleTemplateCache (options, str) {
...
   if (options.compileDebug !== false) {
     fn = [
-        'var jade_debug = [ new jade.DebugItem( 1, ' + filename + ' ) ];'
+        'var jade_debug_filename, jade_debug_line;'
       , 'try {'
       , parsed.body
       , '} catch (err) {'
-      , '  jade.rethrow(err, jade_debug[0].filename, jade_debug[0].lineno' + (options.compileDebug === true ? ',' + utils.stringify(str) : '') + ');'
+      , '  jade.rethrow(err, jade_debug_filename, jade_debug_line' + (options.compileDebug === true ? ', jade_debug_sources[jade_debug_filename]' : '') + ');'
       , '}'
```
alubbe: this is much better - it always bugged me why we store this big array only to use its first member

alubbe:
I've left some comments here and on the commits directly, but it LGTM from what I can tell here. I understand that you also made some changes to the files you extracted (data attributes in double quotes) - this I didn't take a look at but the passing tests should be confirmation enough.

The runtime checking of the for loop is actually not that expensive, so it's further down on my list of improvements. Just to check, are you suggesting to use the ES6 ```for ... of``` loop as the compile target of a jade ```each ... of``` or to compile it to ```for(i;i<l;i++)```?

The big loss of performance actually stems from the mixins and the for loop's IIFE, see https://github.com/jadejs/jade/issues/1975
I'll play around with some ideas of mine and report back :)

PR #1973, opened 2015-06-01
fix documentation, see https://github.com/jadejs/jade/issues/1970

alubbe:
I know we don't want PRs for this, but my reason is to ask how to handle small doc updates like this.
Should I just re-run the script? Should I go directly into gh-pages and update manually?

Maybe we can write down our findings into a Contributing.md. I'm willing to get it started :)

### DIFF
--- a/docs/views/reference/code.jade
+++ b/docs/views/reference/code.jade
@@ -93,7 +93,7 @@ block documentation
     .col-lg-6
       +jade
         :jadesrc
-          p!= 'This code is <strong>not</strong> escaped!'
+          p!= 'This code is' + ' <strong>not</strong> escaped!'
     .col-lg-6
       +html
         :htmlsrc


TimothyGu:
> Should I just re-run the script? Should I go directly into gh-pages and update manually?

No, we usually rerun the script when a release is made.

> Maybe we can write down our findings into a Contributing.md. I'm willing to get it started :)

That would be great! (Also there might be an issue related to this)

alubbe:
So we don't "fix" the docs/the website until the next release? I think it would be great to have a way of updating a specific version's docs even after its release. For example, the 1.10.0 release is still missing the new transformer options.

TimothyGu:
> So we don't "fix" the docs/the website until the next release?

The rule is not "hard" so if you see a need for the fix, sure! I was just saying that we don't usually do this for convenience's sake.

TimothyGu:
LGTM BTW.

alubbe:
I'll merge and publish. Proposal for our policies is at https://github.com/jadejs/jade/pull/1981

@ForbesLindesay Could I interest you in documenting the new transformer options? :)

PR #1974, opened 2015-06-01
removed '=' from title tag in readme

rebeccacaroline:
The "=" sign after title causes an error: unexpected identifier at Function (native) at assertExpression. I removed the "=" and the error was gone. I'm adding screenshots to show my experience.

screen shot 2015-06-01 at 12 46 40 pm
screen shot 2015-06-01 at 12 46 54 pm
screen shot 2015-06-01 at 12 46 59 pm
screen shot 2015-06-01 at 12 47 07 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Jade is a clean, whitespace sensitive syntax for writing html.  Here is a simple
 doctype html
 html(lang="en")
   head
-    title= pageTitle
+    title pageTitle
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
Sorry for the confusion.  The `=` in that example is meant to indicate that the right hand side is a JavaScript expression, rather than literal text.  Perhaps we could make this more clear in the example.

PR #1976, opened 2015-06-02
improved speed of jade.escape()

alubbe:
see https://github.com/jadejs/jade/issues/1975

### DIFF
--- a/lib/runtime.js
+++ b/lib/runtime.js
@@ -179,12 +179,18 @@ exports.attrs = function attrs(obj, terse){
  * @api private
  */
 
+var ENCODE_HTML_RULES = {
+      '&': '&amp;'
+    , '<': '&lt;'
+    , '>': '&gt;'
+    , '"': '&quot;'
+    }
+  , MATCH_HTML = /[&<>"]/g;
+
 exports.escape = function escape(html){
-  var result = String(html)
-    .replace(/&/g, '&amp;')
-    .replace(/</g, '&lt;')
-    .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;');
+  var result = String(html).replace(MATCH_HTML, function(m) {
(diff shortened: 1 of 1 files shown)


lib/runtime.js
```diff
@@ -179,12 +179,18 @@ exports.attrs = function attrs(obj, terse){
...
  */
 
+var _ENCODE_HTML_RULES = {
+      '&': '&amp;'
+    , '<': '&lt;'
+    , '>': '&gt;'
+    , '"': '&quot;'
+    }
+  , _MATCH_HTML = /[&<>"]/g;
+
```
ForbesLindesay: These constants don't need the preceding `_`

ForbesLindesay:
Looks good.  It will also need to be added to the jade-runtime module for 2.0.0

TimothyGu:
LGTM.

Note that sometimes the performance of these escape functions can vary a lot from how you call them. For example, a while ago while tuning ejs I found out that this one you used fares a lot better on client mode (IIRC), while the original is better when you are just compiling and executing in the same environment. That's something to keep in mind when making performance changes.

You should also check if aliasing the variables in the `.replace()` callback and/or `escape()` would help.

alubbe:
@TimothyGu I use both jsperf for the stuff that comes out of the client compilation as well as the server-side benchmark that I forked from yours - so far, I see much better performance in both scenarios and no performance regressions. Do you have a specific benchmark I should look at? 
Could you also help me out in better understanding what you mean by aliasing the variables?

alubbe:
@ForbesLindesay is the jade runtime already extracted into its own repo? The one in this org seems to be a little outdated.

ForbesLindesay:
https://github.com/jadejs/jade-runtime is the 2.0.0 runtime.  It is ahead of 1.0.0 runtime in a few key areas.  If it's behind in anything then that's most likely a mistake.

PR #1983, opened 2015-06-06
for loops are now generated inline rather than in an IIFE

alubbe:
This again improves performance a little bit (part of https://github.com/jadejs/jade/issues/1975).

The first change is to rename ```$index``` to ```$$iN```, where N is the Nth for loop. The PR assumes this is safe. It also assumes it is safe to use variables that look like ```KEY__obj``` and ```KEY__l```, where KEY is either ```$$iN``` or the KEY provided by the user.

### DIFF
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -16,6 +16,7 @@ var Lexer = module.exports = function Lexer(str, filename) {
   this.input = str.replace(/\r\n|\r/g, '\n');
   this.filename = filename;
   this.deferredTokens = [];
+  this.eachCount = 0;
   this.lastIndents = 0;
   this.lineno = 1;
   this.stash = [];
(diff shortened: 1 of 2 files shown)


lib/lexer.js
```diff
@@ -561,7 +562,8 @@ Lexer.prototype = {
     if (captures = /^(?:- *)?(?:each|for) +([a-zA-Z_$][\w$]*)(?: *, *([a-zA-Z_$][\w$]*))? * in *([^\n]+)/.exec(this.input)) {
       this.consume(captures[0].length);
       var tok = this.tok('each', captures[1]);
-      tok.key = captures[2] || '$index';
+      tok.key = captures[2] || '$$i' + this.eachCount;
+      this.eachCount++;
```
alubbe: this is split into two lines on purpose for readability

ForbesLindesay:
This is great, but definitely a breaking change.  Could you make this pull request against [jade-code-gen](https://github.com/jadejs/jade-code-gen) project as well.

alubbe:
Moved variable name generation out of the lexer.
As a side effect, all variable names now start with ```jade_```, so if we consider those safe, the PR is now backwards-compatible.

ForbesLindesay:
Unfortunately I still don't consider this backwards compatible because:

1. The immediately invoked function expression changes the scope of any variables the user defines within it.
2. The `$index` variable that was previously implicitly created, could be used by users.  I'm happy to break these users as part of the 2.0.0 release, but I don't want to break them as a point release.

I'm very excited for getting this done in the 2.0.0 release though :)

alubbe:
Now that we have started landing breaking changes in master, can we merge this or should it go straight into jade-code-gen?

ForbesLindesay:
It needs to go straight into jade-code-gen

PR #1992, opened 2015-06-12
Remove pre-built binaries

ForbesLindesay:
These do not really belong anywhere, least of all the GitHub repo.  We
can generate them as part of the website build and host them elsewhere.


alubbe:
Someone tripped over this: https://github.com/jadejs/jade/issues/2003
Seeing as we currently link to them as part of our README, I would like to generate them for the website or, as a quick fix, update the docs to refer to the files at the 1.11.0 tag - thoughts?

TimothyGu:
> Seeing as we currently link to them as part of our README, I would like to generate them for the website or, as a quick fix, update the docs to refer to the files at the 1.11.0 tag - thoughts?

For now it looks a lot easier to just generate the files one more time. I understand we are going to remove the files eventually anyway, but we can update README then. (npm caches the README from the latest published version so if we fix the docs we'll have to publish a 1.11.1)

But to be honest either one of those solutions is fine for me, as long as the dead link isn't present anywhere.

ForbesLindesay:
Just update the readme. I doubt that many people will be reading the readme on npm and then not using browserify.  We need to keep master moving **forwards** or progress will stall, so regenerating them would be a bad idea.

alubbe:
fixed with https://github.com/jadejs/jade/commit/605aaeae19d1f45a928b87783bca845d818c4cc1

ForbesLindesay:
:+1:

TimothyGu:
+1

PR #1993, opened 2015-06-12
Remove CLI from core jade module

ForbesLindesay:


alubbe:
Just something I noticed - with all of the removals, we basically cannot release another version based on the master branch. Which is fine if tidying things up for 2.0.0 doesn't take long, but could turn into a cherry picking fest on a 1.11.0 based branch if we run into issues.

@ForbesLindesay do you have any estimate of how much time/effort is required to release 2? Are these open issues part of it? https://github.com/jadejs/jade/milestones/2.0.0
What are the most critical or time-consuming parts where we can help out?

ForbesLindesay:
I've updated the roadmap with all the details.  I've also limited that discussion to collaborators to try and minimise the +1/-1 type comments and keep it fairly high signal to noise ratio.

As soon as we landed a single breaking change in master we lost our ability to release a 1.x.x based on it.

PR #1996, opened 2015-06-12
Add JSTransformers reference and usage to Filters docs

markpalfreeman:


alubbe:
I think we could improve this further by linking directly to https://github.com/jstransformers/list-of-jstransformers/blob/master/list-of-jstransformers.json and mentioning the most popular choices, so that most readers can just copy&paste. My vote would be for coffee-script, babel, uglify-js, less and markdown.

TimothyGu:
Keep in mind that people are free to write their own JSTransformers without moving the repo under @JSTransformers. My vote would be to https://www.npmjs.com/search?q=jstransformers

alubbe:
@markpalfreeman it would be awesome if you could update your PR with our input

markpalfreeman:
@alubbe I agree—I like the npm site, and then I'll add in your "frequently used" recommendations.

docs/views/reference/filters.jade
```diff
@@ -4,8 +4,11 @@ block documentation
   h2 Filters
 
   p.
-    Filters let you use other languages within a jade template.
+    Filters let you use other languages within a jade template.
```
alubbe: Do we need that space? ;)

alubbe:
LGTM

docs/views/reference/filters.jade
```diff
@@ -4,8 +4,11 @@ block documentation
   h2 Filters
 
   p.
-    Filters let you use other languages within a jade template.
+    Filters let you use other languages within a jade template. 
     They take a block of plain text as an input.
+  p.
+    All #[a(href='//www.npmjs.com/browse/keyword/jstransformer' target='_blank') JSTransformers] can be used as jade filters.
```
ForbesLindesay: Please use "https://" url here.  Protocol relative urls are an anti-pattern.  Please also remove `target='_blank'`.  If users want to open in a new tab they can do that by holding down CMD.

docs/views/reference/filters.jade
```diff
@@ -4,8 +4,11 @@ block documentation
   h2 Filters
 
   p.
-    Filters let you use other languages within a jade template.
+    Filters let you use other languages within a jade template. 
     They take a block of plain text as an input.
+  p.
+    All #[a(href='//www.npmjs.com/browse/keyword/jstransformer' target='_blank') JSTransformers] can be used as jade filters. 
+    Popular filters include #[code :coffee-script], #[code :babel], #[code :uglify-js], #[code :less], and #[code :markdown].
```
ForbesLindesay: We should encourage `markdown-it` instead of `markdown`.  The `markdown` package on npm is not being actively maintained.

ForbesLindesay:
Looks good apart from those nits.

markpalfreeman:
Great, changes made. Thanks!

PR #2032, opened 2015-08-03
Remove unused dependencies

TimothyGu:


alubbe:
LGTM. Do we still need all of those devDeps?

TimothyGu:
Nope. Removed six that we don't use. Thanks for catching that.

PR #2042, opened 2015-08-12
test: Write to output file before trying to read HTML

TimothyGu:


alubbe:
What's the motivation/rationale?

TimothyGu:
If a test is new, we will be able to directly copy the test/output/blah.html to test/cases/ instead of the `readFileSync` blowing up before the rendering is done.

alubbe:
Okay.
Maybe move the var declaration down as well.
Otherwise, LGTM
Am 12.08.2015 16:54 schrieb "Timothy Gu" <notifications@github.com>:

TimothyGu:
> Maybe move the var declaration down as well.

Not sure why I hadn't done that...

TimothyGu:
Merged in 477e54901dcee8737d94c885110161e75dc5e13e

Issue #1954, opened 2015-05-16
Unbuffered comments are treated as block...


flynx:
**Here is a trivial set of code...**
```jade
mixin foo
  if block
    span there is a block block
    block
  else
    span no block present.
  
+foo
  //- unbuffered comment...
  //- some code we are not using at the moment...
```

**Output:**
```html
<span>there is a block block</span>
```

**Expected:**
```html
<span>no block present</span>
```

This is likely related to #1944

As before, unbuffered comments are treated as blocks while they should be dropped.

*P.S. Am I the only one writing documentation for Jade templates? ;)*

ForbesLindesay:
I'm not sure that I agree in this case.  I would argue that the only reason to put an unbuffered comment there is to explicitly crate an empty block.

I don't mind too much though and I'd love to hear what other people think?

flynx:
@ForbesLindesay in most languages there is a no-op for this, like python's pass, comments are generally used for commenting/documenting or temporarily "commenting out" code, in either case they should not affect the code semantics and should be safe to remove without any effect to the code...

IMO Jade's buffered comments are an exception to the above as they are not really Jade comments but rather HTML comments encoded to Jade and thus significant. unbuffered comments on the other hand, as comments, should be safe to remove without changing the code semantics... and that last bit is currently not true, hence issue ;)

jeromew:
I probably would not have expected the unbuffered comment to create a block but I have never used such orphan comments.

If I wanted to created an empty block, I maybe would have written `= ""` . There are maybe other options to create an empty block.

no sure though if it is worth modifying + it would be a breaking change.

alubbe:
I agree that this is an unexpected behaviour but seems to be a rather rare edge case. Maybe something to look out for in the v2?

flynx:
@alubbe writing documentation for/in templates is a rare edge case? ....this could explain things))))

@jeromew IMHO when comment placement can break code or change it's semantics, it is a problem (see my other issue #1944).

alubbe:
It's not an edge case to write documentation inside a template, but the combination of writing documentation (that does not get rendered into HTML) and checking for content existence via the block attribute seems like an edge case to me. But I may be off.

Anyway, I think that if no actual HTML is rendered, the "block" should be falsy.

flynx:
@alubbe this particular example/issue was a gotcha I got commenting out code while testing, i.e. comments affecting semantics, the more specific documentation case quite common in how I write things is in the other issue (#1944)...

[...]

PR #2043, opened 2015-08-12
Use jade-strip-comments

### LINKED ISSUE #1944: Unbuffered comments can break indented blocks...
**Here is a simple example:**

```jade
if 123
  div some text...
//- comment documenting the else branch...
else
  span some other text...
```

**Result:** this code breaks with the following error pointing to the else keyword:
```
Unexpected else without if
```

**Expected behavior:** unbuffered comments should not affect block handling...

**A bit of motivation:** unbuffered comments are mostly used to document the actual template and are neither needed nor rendered in the HTML, thus should behave as traditional comments (no-op elements) and be simply ignored and dropped by the lexer/translator.

*BTW, this is not an issue for normal "comments" as they translate to HTML comments and their indent is significant and indicates their position in the resulting tree.*

Thanks!

TimothyGu:
Fixes #1944, #1954.

See more discussion at jadejs/jade-parser#8 and jadejs/jade-lexer#7.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,6 +13,7 @@
 var fs = require('fs');
 var path = require('path');
 var lex = require('jade-lexer');
+var stripComments = require('jade-strip-comments');
 var parse = require('jade-parser');
 var load = require('jade-load');
 var filters = require('jade-filters');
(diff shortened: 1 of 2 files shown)


TimothyGu:
Test is failing because jadejs/jade-linker@97bf84037d7ed1c6a07c7267fdf90c72abf274bd hasn't been published.

Issue #1800, opened 2014-12-30
Lines starting with <tabs> <comma> are not parsed correctly


jlhffmn:
In a Jade file indented with tabs, I created an angucomplete-alt directive.  Since it had several attributes, I put a separate attribute on each line.  When I put the comma separators at the beginning of the line the commas were not removed by the Jade conversion to HTML.  

This error did not occur if the commas were at the end of the line or if the tabs were replaced with spaces.

alejonext:
Using no commas
```
.content
    div(angucomplete-alt
        placeholder="Material auto-complete"
        pause="500"
        selected-object="MaterialInfo.Material"
        local-data="MaterialInfo.materialList"
        search-fields="key,description"
        title-field="key,description"
        match-class="angucomplete-highlight"
        minlength="3")
```

PR #2047, opened 2015-08-12
Add test cases for multiline attributes

TimothyGu:
See #1800.


TimothyGu:
See jadejs/jade-lexer#10 for a fix of this case.

alubbe:
Am I right in understanding that the `,` is optional? Basically any combination of `,`, ` ` and `\t` is counted as separators, right? If so, I think it would be a good idea to add a test case without commas.

TimothyGu:
@alubbe done.

PR #2048, opened 2015-08-13
RFC: Expressions in Statements

martindale:
To support a project that is currently locked at 1.7.0, I needed the ability to dynamically evaluate the parameter passed to the `extends` statement.  I quickly hacked in the attached change, which is obviously suboptimal, and am opening this pull request to seek guidance on how this might otherwise be achieved (especially in newer versions of Jade).

What this looks like in practice:
```jade
extends {process.env.PWD + '/views/layouts/default'}

block content
  #page-content !{content}
```

This pull request was implemented on top of the `1.7.0` tag, so it obviously is not mergeable against `master`.  Seeking comments on how this might be addressed otherwise.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -479,8 +479,16 @@ Parser.prototype = {
 
   parseExtends: function(){
     var fs = require('fs');
+    var path = this.expect('extends').val.trim();
 
-    var path = this.resolvePath(this.expect('extends').val.trim(), 'extends');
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
This only works for **Global Variables**, which means for all intents and purposes, it only works for process.env.  That is not useful enough to merit complicating the entire language.

As for solving the more general case, search for "Dynamic Extends" in the issue tracker.

Issue #1871, opened 2015-02-17
Mismatched brackets work with interpolation


TimothyGu:
This should be an error:

```jade
p #[code= 'this is a piece of code'}
```

PR #2049, opened 2015-08-13
Add test cases for inline tags and #1871

TimothyGu:
The test cases are fixed by jadejs/jade-lexer#11.


alubbe:
Do we need to update the html files, as well?

TimothyGu:
> Do we need to update the html files, as well?

I did for the valid case. The anti-case doesn't have an HTML file.

PR #2076, opened 2015-09-07
Added filter options syntax to docs.

geon:
I had to dig through the parser/lexer source to find out that this was actually possible and supported. It's really a shame that such an important feature is not undocumented.


TimothyGu:
> I had to dig through the parser/lexer source to find out that this was actually possible and supported

Sorry about that. Next time you could try looking at the test cases. For example, https://github.com/jadejs/jade/blob/master/test/cases/filters.coffeescript.jade has `:coffee-script(minify=true)`.

PR #2092, opened 2015-09-17
docs/client: Force external runtime

TimothyGu:
jade-runtime/build doesn't work when Browserified.


TimothyGu:
jade-runtime/build has been fixed: jadejs/jade-runtime#22. Patch dropped.

PR #2093, opened 2015-09-17
Document support for template strings in attributes 

### LINKED ISSUE #1447: Document support for multi-line strings
The documentation doesn't cover this, but I've found that ending a line with a back slash allows a string to continue onto the next line, which is something you'd expect.

```
div(attr="foo \
  bar \
  baz")
```

compiles to

```
<div attr="foo   bar   baz"></div>
```

It would be nice if this was documented.

TimothyGu:
Fixes #1447.

### DIFF
--- a/docs/client/index.js
+++ b/docs/client/index.js
@@ -53,7 +53,29 @@ function cm(el, mode, readonly) {
   }
 }
 
+var features = {
+  'templatestrings': (function () {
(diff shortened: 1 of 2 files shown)


alubbe:
What do you think about adding a variable into the string, as well?

TimothyGu:
Done.

Issue #2082, opened 2015-09-12
Self Closing Tags should be respected in HTML, especially in SVG


ryanvfung:
I am using Jade v1.11.0 through Node.js
While the DOCTYPE specified is HTML, Jade should still respect self closing tags (e.g.: `<path/>`, `<rect/>`, `<circle/>`, etc.)  especially in embedded XML types such as SVG (maybe also MathML).

Related to #1510

Example Jade input:
```jade
doctype html
html
  head
  body
    svg(width="200px", height="230px", version="1.1", xmlns="http://www.w3.org/2000/svg", xmlns:xlink="http://www.w3.org/1999/xlink")
      <circle cx="100" cy="100" r="100" fill="#000" />
      <circle cx="100" cy="100" r="95" fill="#ff0" />
    svg(width="200px", height="230px", version="1.1", xmlns="http://www.w3.org/2000/svg", xmlns:xlink="http://www.w3.org/1999/xlink")
      circle/(cx="100", cy="100", r="100", fill="#000")
      circle/(cx="100", cy="100", r="95", fill="#ff0")
```
produces the following HTML output:
```html
<!DOCTYPE html>
<html>
  <head></head>
  <body>
    <svg width="200px" height="230px" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
      <circle cx="100" cy="100" r="100" fill="#000" />
      <circle cx="100" cy="100" r="95" fill="#ff0" />
    </svg>
    <svg width="200px" height="230px" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
      <circle cx="100" cy="100" r="100" fill="#000">
      <circle cx="100" cy="100" r="95" fill="#ff0">
    </svg>
  </body>
</html>
```
When rendered, the following is shown:
![2015-09-12 23-05-41](https://cloud.githubusercontent.com/assets/2174715/9834038/dd8e26da-59a2-11e5-833a-51496b1ed463.png)

As shown by the example render, the output is interpreted differently.
The produced HTML looks almost identical, but because the circle tag is not closed, parsers such as Google Chrome will keep them open and the second circle tag is a child of the first circle tag, causing incorrect / unexpected results.
Both SVG tag contents should have identical Jade outputs if self closing tags are respected.

TimothyGu:
PR available at jadejs/jade-code-gen#15.

PR #2102, opened 2015-09-22
Docu typos

Laoujin:
While doing the exercises from http://jade-lang.com/reference/, I stubled upon what I think are some small typos.

Also not sure about this one but, http://jade-lang.com/reference/tags/ mentions at the end "Self Closing Tags". But doing `foo/`generates a `<foo>` and not a `<foo />` as advertised. Not sure if this is a bug.

### DIFF
--- a/docs/views/reference/comments.jade
+++ b/docs/views/reference/comments.jade
@@ -70,15 +70,15 @@ block documentation
           <!--[if IE 8]>
           <html lang="en" class="lt-ie9">
           <![endif]-->
-          <!--[if gt IE 8]><!-->
+          <!--[if gt IE 8]>
           <html lang="en">
-          <!--<![endif]-->
+          <![endif]-->
     .col-lg-6
       +html
         :htmlsrc
           <!--[if IE 8]>
           <html lang="en" class="lt-ie9">
           <![endif]-->
-          <!--[if gt IE 8]><!-->
+          <!--[if gt IE 8]>
           <html lang="en">
-          <!--<![endif]-->
+          <![endif]-->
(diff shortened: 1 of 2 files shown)


TimothyGu:
Both of them are intentional.

For comments, the extra comment tags are required to make browsers other than IE to recognize the body tag. See http://stackoverflow.com/a/9254532/1937836 for an example.

For inheritance, it is intentional as well. `pet.jade` is included repeatedly for each pet.

TimothyGu:
> Also not sure about this one but, http://jade-lang.com/reference/tags/ mentions at the end "Self Closing Tags". But doing `foo/`generates a `<foo>` and not a `<foo />` as advertised. Not sure if this is a bug.

It is a bug: #2082.

Laoujin:
```
block content
  h1= title
  each pet in pets
    include pet
```

This had me confused. I thought I could dynamically include jade that way, but after some testing it turned out this was not possible.

It doesn't make sense to include the same jade for every pet. I would still suggest to change this in order to avoid this confusion for other new Jaders.

TimothyGu:
> It doesn't make sense to include the same jade for every pet.

The inclusion is used as if it were a mixin, without the extra overhead of method call during runtime. I'd say that it does make sense.

Laoujin:
Perhaps I'm missing something here...

Is this equivalent?
```
each petName in ['cat', 'dog']
  include pet.jade
```

Where I assumed it would be:
```
include cat.jade
include dog.jade
```

TimothyGu:
No.

```jade
//- pet.jade
p= petName
```

```jade
//- main.jade
each petName in ['cat', 'dog']
  include pet.jade
```

is equivalent to

```jade
each petName in ['cat', 'dog']
  p= petName
```

PR #2103, opened 2015-09-22
Documentation-Reference: Attributes: Comma is optional

Laoujin:
This adds the documentation request by: https://github.com/jadejs/jade/issues/2029

(not sure if it's good enough)

### DIFF
--- a/docs/views/reference/attributes.jade
+++ b/docs/views/reference/attributes.jade
@@ -3,18 +3,20 @@ extends ../reference.jade
 block documentation
   h1 Attributes
 
-  p Tag attributes look similar to html, however their values are just regular JavaScript.
+  p Tag attributes look similar to html (with optional comma),
+    | however their values are just regular JavaScript.
 
   .row(data-control='interactive')
     .col-lg-6
       +jade
         :jadesrc
-          a(href='google.com') Google
+          a(class='button' href='google.com') Google
           a(class='button', href='google.com') Google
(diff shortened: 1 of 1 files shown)


docs/views/reference/attributes.jade
```diff
@@ -3,18 +3,20 @@ extends ../reference.jade
 block documentation
   h1 Attributes
 
-  p Tag attributes look similar to html, however their values are just regular JavaScript.
+  p Tag attributes look similar to html (with optional comma),
+    | however their values are just regular JavaScript.
```
TimothyGu: Please merge these two lines.

docs/views/reference/attributes.jade
```diff
@@ -3,18 +3,20 @@ extends ../reference.jade
...
         :jadesrc
-          a(href='google.com') Google
+          a(class='button' href='google.com') Google
           a(class='button', href='google.com') Google
     .col-lg-6
       +html
         :htmlsrc
-          <a href="google.com">Google</a><a class="button" href="google.com">Google</a>
+          <a class="button" href="google.com">Google</a>
+          <a class="button" href="google.com">Google</a>
```
TimothyGu: For future reference, always use the output from Jade (with whitespace preserved) in `:htmlsrc` blocks of the demos. In this case, Jade does not output the new lines in between the two tags and the new lines here are confusing.

PR #2104, opened 2015-09-22
Documentation-Reference: Inheritance: 

Laoujin:
Additional context for `include pet` to avoid confusion for those that didn't figure out the variable scoping - like me :)


TimothyGu:
Thanks!

PR #2132, opened 2015-10-16
Add missing `-H, --hierarchy` flag to command-line template.

akivajgordon:
The `-H` flag appears for

```
$ jade --help
```

but not at http://jade-lang.com/command-line/


TimothyGu:
This flag will be removed in the next release of jade-cli (which will be at the same time as the next release of jade), so this PR wouldn't be necessary.

akivajgordon:
@TimothyGu Ok, no problem. Can you offer any insight as to why this option is being removed?

TimothyGu:
The option is made default, and the old behavior was removed.

PR #2167, opened 2015-11-16
Update uglify.js to 2.6.1

ockham:
To include a fix (mishoo/UglifyJS2@63d35f8) for https://nodesecurity.io/advisories/48

Also filed jstransformers/jstransformer-uglify-js#6 to update that dependency.


hemanth:
Thanks!

PR #2169, opened 2015-11-18
Update jstransformer-uglify-js dependency to 1.1.1

ockham:
To include the updated `uglify.js` dependency fix in `jstransformer-uglify-js` (jstransformers/jstransformer-uglify-js@e897c68)

Please also do a point release!


TimothyGu:
I updated jade-filters but not jade… thanks!

PR #2171, opened 2015-11-20
Add plugin infrastructure

TimothyGu:
This is a work-in-progress. Two plugins I have made are available here:

- [jadejs/jade-plugin-debug-line](https://github.com/jadejs/jade-plugin-debug-line)
- [jadejs/jade-plugin-dynamic-include](https://github.com/jadejs/jade-plugin-dynamic-include)

To use these two plugins, first clone them. Assuming that they are cloned in directories parallel to the `jade` directory (which of course has this `plugins` branch checked out), and assuming that you are in the `jade` directory, try the following.

```js
var fs = require('fs');
var jade = require('.');
var str;

// sources
fs.writeFileSync('my-jade-src.jade', 'p hey');
fs.writeFileSync('interesting-styles.css', 'body { background-color: black; color: white }');
fs.writeFileSync('includer.jade', "- var file = 'interesting-styles'\nstyle\n  dyninclude #{file}.css\n- file = 'my-jade-src'\ndyninclude #{file}.jade");


// debug-line
str = jade.renderFile('my-jade-src.jade', {
  plugins: ['../jade-plugin-debug-line'] // once the module is published one can simply use `['debug-line']`
});
console.log(str);
// <p data-jade-file="my-jade-src.jade" data-jade-line="1">hey</p>


// dynamic-include
str = jade.renderFile('includer.jade', {
  plugins: [
    '../jade-plugin-dynamic-include'
  ],
  // unfortunately `require` is not available inside a dynamically-generated function, so have to pass in these two as locals
  fs,
  jadeSelf: jade
});
console.log(str);
// <style>body { background-color: black; color: white }</style><p>hey</p>
```

Unimplemented hooks:

- linker plugins
- code generator plugins

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -12,6 +12,7 @@
 
 var fs = require('fs');
 var path = require('path');
+var resolve = require('resolve').sync;
 var lex = require('jade-lexer');
 var stripComments = require('jade-strip-comments');
 var parse = require('jade-parser');
@@ -52,14 +53,60 @@ function compileBody(str, options){
   var debug_sources = {};
   debug_sources[options.filename] = str;
   var dependencies = [];
+  var plugins = options.plugins || [];
+  plugins = plugins.map(function (plugin) {
+    if (typeof plugin === 'string') {
+      var opts = {
+        basedir: path.dirname(module.parent.filename)
+      };
+      // Test if plugin name is local (i.e. `./my-plugin`)
+      // Regex from https://github.com/substack/node-resolve/blob/5cae82fb22cb64d5b72f703c787dc0fd418ed412/lib/sync.js#L21
+      if (/^(?:\.\.?(?:\/|$)|\/|([A-Za-z]:)?[\\\/])/.test(plugin)) {
+        return require(resolve(plugin, opts))();
+      } else {
+        return require(resolve('jade-plugin-' + plugin, opts))();
+      }
+    }
+    return plugin;
+  });
+
   try {
     var ast = load.string(str, options.filename, {
-      lex: lex,
+      lex: function (str, filename) {
+        getPlugin('preLex').forEach(function (func) {
+          str = func(str);
+        });
+        var tokens = lex(str, filename, {
+          plugins: getPlugin('lex')
+        });
+        getPlugin('postLex').forEach(function (func) {
+          tokens = func(tokens);
+        });
+        return tokens;
+      },
       parse: function (tokens, filename) {
+        getPlugin('preParse').forEach(function (func) {
+          tokens = func(tokens);
+        });
         tokens = stripComments(tokens, { filename: filename });
-        return parse(tokens, filename);
+        var ast = parse(tokens, filename, {
+          plugins: getPlugin('parse')
+        });
+        getPlugin('postParse').forEach(function (func) {
+          ast = func(ast);
+        });
+
+        getPlugin('preLoad').forEach(function (func) {
+          ast = func(ast);
+        });
+        return ast;
       },
       resolve: function (filename, source) {
+        var plugins = getPlugin('loadResolve');
+        for (var i = 0; i < plugins.length; i++) {
+          var result = plugins[i](filename, source, options);
+          if (result) return result;
+        }
         filename = filename.trim();
         if (filename[0] !== '/' && !source)
           throw new Error('the "filename" option is required to use includes and extends with "relative" paths');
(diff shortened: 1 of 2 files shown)


TimothyGu:
@ForbesLindesay ping. We'll need (a lot) more tests for this but the feature itself looks okay, but I'm totally not biased :p

lib/index.js
```diff
@@ -52,14 +53,60 @@ function compileBody(str, options){
...
+  var plugins = options.plugins || [];
+  plugins = plugins.map(function (plugin) {
+    if (typeof plugin === 'string') {
+      var opts = {
+        basedir: path.dirname(module.parent.filename)
+      };
+      // Test if plugin name is local (i.e. `./my-plugin`)
+      // Regex from https://github.com/substack/node-resolve/blob/5cae82fb22cb64d5b72f703c787dc0fd418ed412/lib/sync.js#L21
+      if (/^(?:\.\.?(?:\/|$)|\/|([A-Za-z]:)?[\\\/])/.test(plugin)) {
+        return require(resolve(plugin, opts))();
```
ForbesLindesay: This seems a little bit brittle.  What if multiple modules call `require('pug')`?  I think it might be better to require absolute paths for local modules (or paths relative to current working directory).

lib/index.js
```diff
@@ -52,14 +53,60 @@ function compileBody(str, options){
...
+      } else {
+        return require(resolve('jade-plugin-' + plugin, opts))();
```
ForbesLindesay: This looks good here.  We should use the current working directory when called from the jade cli.

lib/index.js
```diff
@@ -112,6 +187,16 @@ function compileBody(str, options){
     }
     throw err;
   }
+
+  function getPlugin(hook) {
```
ForbesLindesay: Rename this to `getPlugins` since it returns an array.

lib/index.js
```diff
@@ -52,14 +53,60 @@ function compileBody(str, options){
...
+          ast = func(ast);
+        });
+
+        getPlugin('preLoad').forEach(function (func) {
+          ast = func(ast);
+        });
+        return ast;
       },
       resolve: function (filename, source) {
+        var plugins = getPlugin('loadResolve');
```
ForbesLindesay: Should the `hook` be `loadResolve` or should it be `getPlugin('load').map(loaderPlugin => loaderPlugin.resolve).filter(Boolean)`?

lib/index.js
```diff
@@ -52,14 +53,60 @@ function compileBody(str, options){
...
+        for (var i = 0; i < plugins.length; i++) {
+          var result = plugins[i](filename, source, options);
+          if (result) return result;
```
ForbesLindesay: The empty string should be considered a valid file path here.  Lets do `result !== undefined` as our test.

lib/index.js
```diff
@@ -74,15 +121,40 @@ function compileBody(str, options){
         return filename;
       },
       read: function (filename) {
+        var str;
+        var plugins = getPlugin('loadRead');
+        for (var i = 0; i < plugins.length; i++) {
+          str = plugins[i](filename, source, options, dependencies);
+          if (str) {
```
ForbesLindesay: The empty string should be considered file content here.  Lets do `str !== undefined` as our test.

lib/index.js
```diff
@@ -74,15 +121,40 @@ function compileBody(str, options){
...
+        str = fs.readFileSync(filename, 'utf8');
         debug_sources[filename] = str;
         return str;
       }
     });
+    getPlugin('postLoad').forEach(function (func) {
+      ast = func(ast);
+    });
+
     ast = filters.handleFilters(ast, exports.filters);
```
ForbesLindesay: Unrelated to this change, but we should merge `options.filters` with `exports.filters` so that you can do `jade.compile(str, {filters: {'my-custom-filter': function (str) { ... }}})`

lib/index.js
```diff
@@ -74,15 +121,40 @@ function compileBody(str, options){
...
-    ast = link(ast);
 
+    getPlugin('preLink').forEach(function (func) {
+      ast = func(ast);
+    });
+    ast = link(ast, {
+      plugins: plugins
```
ForbesLindesay: Should be `plugins: getPlugins('link')`

lib/index.js
```diff
@@ -74,15 +121,40 @@ function compileBody(str, options){
...
+        var str;
+        var plugins = getPlugin('loadRead');
+        for (var i = 0; i < plugins.length; i++) {
+          str = plugins[i](filename, source, options, dependencies);
+          if (str) {
+            debug_sources[filename] = str;
+            return str;
+          }
+        }
         dependencies.push(filename);
```
ForbesLindesay: We need to allow the plugin some way to **optionally** indicate a dependency on a file.

ForbesLindesay:
Looks really good.  One thought:  In all the pre/post hooks, lets pass the filename, basedir etc. as a second argument.  We should also probably provide them with methods for pushing into the array of `dependencies` and the `debug_sources`.

Also, lets provide an API of some kind that allows a single plugin to replace any of the stages.  Not sure what that should look like.

PR #2186, opened 2015-12-12
update travis-ci to use pugjs instead of jadejs

OmgImAlexis:
Since the repo is at pugjs/jade and not jadejs/jade the travis badge was no longer working.


TimothyGu:
This has become obsolete.

Issue #1927, opened 2015-04-10
Include some documentation on how to build the gh-pages


GarthDB:
I was looking to do a pull request with updates to the jade-lang.com pages and I couldn't find an obvious way to build the gh-pages branch.

It would be nice to have some documentation in case anyone wants to make contributions to the front end design.

I'd be happy to write it up if someone could point me in the right direction.

ForbesLindesay:
This would be great.

The website/documentation lives in the `docs` directory.  To work on the website:

1. Install the latest version of [node](https://nodejs.org/) or [iojs](https://iojs.org/)
2. Create a fork of this repository using the GitHub "Fork" button
3. Clone the repository using git
4. Run `npm install` in a terminal/command line in the root of this repository to fetch dependencies
5. `cd docs` then run `node server` to start a development server for the documentation
6. Open http://localhost:3000 in a web-browser to view your local copy of the website/documentation
7. Make edits in the `/docs` folder of this repository (the `/docs/views` folder contains all the jade files)
8. Push your changes and submit a pull request.
9. One of the jade core team will review your changes, merge them, and ultimately build the static gh-pages version of the site when the next release of jade is published.


Where do you think would be the best location for that documentation?

GarthDB:
That was the first thing I tried and it wasn't working, I just hard reset to master and it fixed it.

We should probably make it a section in a CONTRIBUTING file. It's pretty standard on GitHub.  I'll make a pr on Monday, if you don't get to it.

pixelastic:
I also wanted to contribute to the documentation but can't find how to build it.

- I checked the `gh-pages` of this repo but it seems to be outdated (it's mentionning jade-lang.org)
- I checked https://github.com/pugjs/pug-www but could not find the actual page content, only running what seems like an empty server
- I checked https://github.com/pugjs/pug-en but could not find how to run it
- I also checked https://github.com/pugjs/pug-fr but it seems to be outdated (https://pugjs.org/fr/ does not work)

I'd like to submit a PR to add search to the pug documentation, could you help me pointing me how to build the documentation locally?

ForbesLindesay:
Sorry I didn't see this comment. The current approach to "running" the docs is to clone both pug-en and pug-www into the same parent directory. You then run pug-www and edit pug-en.  In this way, you can also clone any of the pug-(language) repositories to see/edit docs in that language. I'm sorry I haven't been great at making the process smooth lately. It's something I'd love to review if I could find time.

PR #2202, opened 2016-01-05
Class literal is duplicating the bing class

chadfawcett:
In the **Class Attributes** part of the Language Reference, the anchor tag has a **Class Literal** attached. This, in addition to the second class attribute is making it so the element has **bing** as a class twice. Judging by the docs source, this is not the intended outcome as it shows only one **bing** class.

The `/docs` source shows the following:
```html
:htmlsrc
  <a class="foo bar baz"></a><a class="foo bar baz bing"></a>
```

The generated html on the live site:
```html
<a class="foo bar baz"></a><a class="bing foo bar baz bing"></a>
```

**Note:** I wasn't able to get the CodeMirror working when following the contribution instructions found on #1927 so the html source wasn't being generated from the jade source which meant the html source was showing the hard coded value which was the desired result. If the code was to be generated from the jade source, it would have the second **bing** class from the Literal.

### DIFF
--- a/docs/views/reference/attributes.jade
+++ b/docs/views/reference/attributes.jade
@@ -172,11 +172,11 @@ block documentation
           - var classes = ['foo', 'bar', 'baz']
           a(class=classes)
           //- the class attribute may also be repeated to merge arrays
-          a(class=classes class=['bing'])
+          a.bang(class=classes class=['bing'])
     .col-lg-6
       +html
         :htmlsrc
-          <a class="foo bar baz"></a><a class="foo bar baz bing"></a>
+          <a class="foo bar baz"></a><a class="bang foo bar baz bing"></a>
 
   p It can also be an object mapping class names to true or false values, which is useful for applying conditional classes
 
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
We should probably just change one of the two bing's to something else, then include all the correct values in the output.

chadfawcett:
Should we still do that considering the documentation for the class literals isn't until right after the class attribute documentation? I would think it would be more clear to not have the literal in this example.

Let me know and I'll make changes if you'd like.

chadfawcett:
@ForbesLindesay I've changed the class literal to be `bang` so it is now unique.

ForbesLindesay:
Thanks!  Sorry for not getting round to replying.  I'm not sure I have a strong opinion, but this is definitely an improvement over what we had before.

TimothyGu:
I believe that the deduplication is not working is a bug in itself. I'll take a look at it.

chadfawcett:
@TimothyGu I was wondering that as well. Thanks for taking a look.

PR #2219, opened 2016-01-19
Revert "Add test for variable interpolation in HTML"

sg1303:
Fix broken travis build.

### DIFF
--- a/test/cases/html.html
+++ b/test/cases/html.html
@@ -3,7 +3,5 @@
 <li>bar</li>
 <li>baz</li>
 </ul>
-<!--build:js /js/app.min.js?v=1449104952939-->
-<!--endbuild-->
 <p>You can <em>embed</em> html as well.</p>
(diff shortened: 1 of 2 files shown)


sg1303:
Reverted addition of test case that tried to add an interpolated string inside of an html comment, but it's evaluated as plaintext like this says:

> Jade does not have any special syntax for conditional comments. If your line begins with `<` then it is > treated as plain text.

TimothyGu:
A test case this there for a reason. The code should be fixed if the test fails.

sg1303:
I don't understand... sounds like you are proposing a feature be added? In my mind this is currently working as designed.

ForbesLindesay:
We allow interpolation in plain text. We already have `//` for true comments.  HTML comments are intended for things like knockout conditionals, where you would want interpolation to work.

sg1303:
That makes sense. Thanks for clarifying.

Issue #2231, opened 2016-01-30
Delete


niftylettuce:

niftylettuce:
Done!  http://jadelang.net/ and http://jadesoftware.org are now live.  There are 2530 results for 'jade' on NPM.  So they can have fun filing claims against 2530 packages, as opposed to actually making their clients satisfied.  I believe there are a few other domain names like these available, if anyone else wants to purchase and support this anti-troll case then please do. :trollface:

niftylettuce:
https://github.com/jadejs/jade
https://github.com/jadesoftware/jade

Please don't give in to this BS :hankey:.

DiegoRBaquero:
+1 into not giving in to this BS. Their website being down is just the start.

PR #2244, opened 2016-02-15
Change repo links to pugjs/jade

nathanhleung:
Right now jadejs/jade is a different person's repository (squatter maybe?). I changed Travis to pugjs as well, but coveralls gives me a 404 at pugjs/jade.


TimothyGu:
It was a remnant of #2231.

PR #2246, opened 2016-02-15
correctly resolve files in node_modules

### LINKED ISSUE #2164: how can i include some jade from node_modules?
example 

```
header.jade in node_modules/xxx/views/header.jade
```

in my express layout 

```
include node_modules/xxx/views/header.jade
```

tellnes:
Hi

This lets you do `extends foo/bar` which might resolve to `node_modules/foo/bar.jade`. It also keeps backwards compatibility which means it also might match `./foo/bar.jade` which is incompatible with nodes `require` lookup algorithm.

Fixes #2164


ForbesLindesay:
I think for now, this would be better done in a plugin. I also think it might make sense to either:

1. Require a prefix like `npm://package-name` to indicate that `node_modules` is being used.
2. Or deprecate paths that are local, but don't start with `./` (i.e. mostly just start using the node resolution algorithm for everything).

1 would be easy, but is a little bit messy, 2 is a long term thing that I don't think we should add to the already very late 2.0.0 release.

tellnes:
I think 2. is the best option, but maybe we can add 1. to the v1.x.x branch? I don't think a prefix is necessary since the current resolve logic take precedence. The braking changed is that if the current resolution does not resolve, it tries to use node resolution algorithm.

Also, I see that a lot of tests is failing on travis. I'll look into that after a decision for which path to take forward is made.

One question. I'm not very familiar with how the jade development is organized. Is the `v2.0.0` release in the master branch and all `v1.x.x` releases in the `v1.x.x` branch?

ForbesLindesay:
I'm going to close this as it is easy to resolve via a plugin.

PR #2279, opened 2016-03-29
chore(deps): updated transformers because of node security warning

jzsfkzm:
Hey there,

I just updated this, because transformers 2.1.0, the version referenced is pulling uglify-js 2.2.5, and that version of uglify-js has security issues according to nodesecurity:
https://nodesecurity.io/advisories/39

We're still on the 1.x branch of jade, can you please merge and release this as a fix?

Thanks
Jozsef

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "constantinople": "~3.0.1",
     "jstransformer": "0.0.2",
     "mkdirp": "~0.5.0",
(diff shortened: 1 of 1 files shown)


TimothyGu:
This has been proposed before (by me), but since transformers 3 is not fully compatible to 2 we unfortunately can't do that for 1.*.

PR #2321, opened 2016-04-18
Add a test case for template string in attributes

TimothyGu:


ForbesLindesay:
We need to either drop support for 0.10 and 0.12 (probably shouldn't do that until the next major version) or we need to skip this test on those platforms.

TimothyGu:
> We need to either drop support for 0.10 and 0.12 (probably shouldn't do that until the next major version) or we need to skip this test on those platforms.

Ugh I really shouldn't do things while I'm half-asleep…

Issue #2328, opened 2016-04-20
The file extension `.jade` is injected when you run `extends '<my file>'` without an explicit extension.


maraisr:

mccart:
had to specify explicit extension .pug to get it to work.

ForbesLindesay:
Please ensure you have updated to the latest version of pug and try again.  If the problem persists, you can comment here and I will re-open.

mccart:
$ npm update pug 
does nothing...  error still persists.

adding this to package.json
    "pug": "*",
+   "pug": "https://github.com/pugjs/pug.git"

$ npm update
express_website@1.0.0 /home/mccart/Projects/NodeJSProjects/BasicExpressWebsite
└── pug@2.0.0-alpha6  (git+https://github.com/pugjs/pug.git#610972b2cedad7861ed9a447792112559615b4c3)

Success - no longer have to explicitly specify extension .pug for extends <file>

ForbesLindesay:
npm's update command doesn't update the currently installed packages and I'm not sure `*` includes alpha versions.  Set it to `"pug": "2.0.0-alpha6"` then run `npm install`.

Issue #2312, opened 2016-04-15
When using "extends" if not specidfied PUG looks for "jade" extension.


paran0a:
If I have something like 

`extends layout`

I'll get an error

`Error: ENOENT: no such file or directory, open 'C:...\views\layout.jade'`

If I specify that its pug extension like `extends layout.pug` then it works.

jeremyrajan:
👍 , is there a fix for it yet?

ForbesLindesay:
This isn't the behaviour of pug 2.0.0.  I think you're probably using an out of date version.

dcollao:
I have the same error :( Error: ENOENT: no such file or directory, open '/Users/3ndgl0ry/Projects/user-login-system/views/layout.jade'
   at /Users/3ndgl0ry/Projects/user-login-system/views/error.pug line 1
   at Error (native)
   at Object.fs.openSync (fs.js:549:18)
   at Object.fs.readFileSync (fs.js:397:15)
   at Object.load.string.read (/Users/3ndgl0ry/Projects/user-login-system/node_modules/pug/lib/index.js:78:22)
   at Function.read (/Users/3ndgl0ry/Projects/user-login-system/node_modules/pug/node_modules/pug-loader/index.js:66:47)
   at /Users/3ndgl0ry/Projects/user-login-system/node_modules/pug/node_modules/pug-loader/index.js:23:22
   at walkAST (/Users/3ndgl0ry/Projects/user-login-system/node_modules/pug/node_modules/pug-loader/node_modules/pug-walk/index.js:13:26)
   at /Users/3ndgl0ry/Projects/user-login-system/node_modules/pug/node_modules/pug-loader/node_modules/pug-walk/index.js:21:16
   at Array.map (native)
   at walkAST (/Users/3ndgl0ry/Projects/user-login-system/node_modules/pug/node_modules/pug-loader/node_modules/pug-walk/index.js:20:29)

ForbesLindesay:
Please make sure you are using the latest version of pug.

jtmdrums:
Using pug v 3.3.12 on a mac, still getting that error unless I specifically write "extends layout.pug"

ForbesLindesay:
There is no such thing as pug v3.3.12.  The latest version is currently 2.0.0-beta11

Issue #2281, opened 2016-03-30
Stop auto-appending `.pug` to includes/extends


dizzib:
pug v2.0.0-alpha1
pug-cli v1.0.0-alpha1

For example:

foo.pug

    include _bar
    +bar

_bar.pug

    mixin bar
      .bar

...compiling:

    [andy@kango view (pugfoo)]$ ../../../node_modules/.bin/pug ./foo.pug 

    /home/andy/project/wdts/node_modules/pug/node_modules/pug-runtime/index.js:252
      throw err;
            ^
    TypeError: ./foo.pug:3
        1| include _bar
        2| 
      > 3| +bar
        4| 

    Object #<Object> has no method 'bar'
        at template (eval at wrap (/home/andy/project/wdts/node_modules/pug/node_modules/pug-runtime/wrap.js:6:10), <anonymous>:5:18)
        at renderFile (/home/andy/project/wdts/node_modules/pug-cli/index.js:282:40)
        at /home/andy/project/wdts/node_modules/pug-cli/index.js:146:5
        at Array.forEach (native)
        at Object.<anonymous> (/home/andy/project/wdts/node_modules/pug-cli/index.js:145:9)
        at Module._compile (module.js:456:26)
        at Object.Module._extensions..js (module.js:474:10)
        at Module.load (module.js:356:32)
        at Function.Module._load (module.js:312:12)
        at Function.Module.runMain (module.js:497:10)

But if I change foo.pug to include the `.pug` extension...

    include _bar.pug
    +bar

...then compiling works as expected:

    [andy@kango view (pugfoo)]$ ../../../node_modules/.bin/pug ./foo.pug

      rendered foo.html
    [andy@kango view (pugfoo)]$ cat foo.html
    <div class="bar"></div>

This problem does not occur in Jade 1.9.2.

EDIT: Ah, I just saw #2268. In which case shouldn't the error be `ENOENT no such file` since the file `_bar` does not exist ?

ForbesLindesay:
Yes, we were planning to continue appending `.pug` automatically, but in the end we broke support for that when simplifying the distinction between RawIncludes and pug Includes.  We should remove the code that automatically adds `.pug` so as to reduce confusion.

ForbesLindesay:
We now auto-apend .pug still, but it should work properly.

PR #2334, opened 2016-04-26
use native path module to parse include/extends file extension

doublerebel:
Hello,

The [current test for `'.'` in the path](https://github.com/pugjs/pug/commit/433e2ab18b3aabe7b34947489f799c0390aa0ac4) excludes all paths with a `.` including `../`, breaking some relative paths.  I've written tests to demonstrate this case.  First I am pushing the tests which fail, then the fix which passes.

Thanks for your work!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -51,7 +51,7 @@ function compileBody(str, options){
     lex: lex,
     parse: function (tokens, filename) {
       tokens = tokens.map(function (token) {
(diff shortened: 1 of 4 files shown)


doublerebel:
Ok, now that [the test case has failed](https://travis-ci.org/pugjs/pug/builds/125745395), I'll push the fix.  Tagging #2281 #2312 and #2328 in case anyone else besides me is looking for this answer.

ForbesLindesay:
[closes #2336]

PR #2341, opened 2016-04-28
add doclets.io config

lipp:
Adds documentation to [doclets.io](https://doclets.io) service. Future tags and master branch will be published automatically.

[Preview generated with my account](https://doclets.io/lipp/pug/add-to-doclets)

@TimothyGu / @ForbesLindesay   If you like this, you are very welcomed to login to https://doclets.io with GitHub OAuth. If you don't like it, I'd be happy to hear what you don't like about it :)

Steps required (3min):
1) @TimothyGu   logs in to doclets.io
2) @TimothyGu   adds pug
3) someone merges this PR

### DIFF
--- a/.doclets.yml
+++ b/.doclets.yml
@@ -1,0 +1,4 @@
+dir: lib
+articles:
+  - About: README.md
+  - History: History.md


ForbesLindesay:
This does look like a pretty cool service.  I have been considering building something similar to this sort of thing myself.  I don't want to split the docs across two separate websites though, which means that we would only really want to use this if it could replace the website.  I think there are a number of hurdles to doing that:

1. We would want to host it on our own domain to make it easier to migrate in the future if doclets ceases to suit our needs.
2. We would want to present one unified main version, currently the branches are given a lot of prominence, which would be un-necessary confusion for new users.
3. We would want a navigation structure similar to what you get on http://facebook.github.io/react/ (i.e. use the top header for the primary navigation information and have a home page that doesn't have any sidebar for navigation.
4. We would want to be able to customise the colour scheme to match our branding and use our logo in place of the doclet logo.
5. Our examples in our documentation are all interactive.  We would need to be able to embed our own custom JS widgets somehow within the documentation.
6. I expect we will be looking at translating the website into multiple languages fairly soon.  We would want whatever tool we use to provide good support for translations.

lipp:
@ForbesLindesay Thanks a lot for your thorough feedback 🎱 ! This is worth very much to me! I hope that I can address your requirements in the future. Some remarks to your concerns:

1. This is a point BIG projects like pug will have in most cases I guess. You have a comprehensive project page in place and you don't want to redirect to (several) sub services/pages. Anyhow, I don't know yet how to realize this... Doclets could push/upload the doc pages e.g. to the gh-pages. Would this be ok?
2. That is a very interesting observation. Didn't saw it that way yet. I'll try to set the focus more on the "latest" doc.
3. The layout / structure is pretty basic right now. I want to add a (NICE) search field (autocomplete etc) and other dropdown menus for quicker navigation. I am not very convinced by the sidebar tree view either.
4. This shouldn't be to hard to achieve. Would setting this through the .doclets.yml config file ok for you?
5. You can actually use tonic dev interactive editor right now using the non-standard "@playground" tag: https://doclets.io/lipp/acme-jsdoc-example/master#dl-greet . Users were already asking for something like jsFiddle support. Would you still need more customizable stuff? If so, how would you like to embed it?
6. Don't know how to do this with JSDoc... maybe there is a plugin available...

Thanks again. Even if you won't use it for pug, I'd be supper happy if you may spread the word 💯

ForbesLindesay:
1. Pushing to something like gh-pages would be fine.  My plan is to upload to Amazon S3 and host there, but I can do that from gh via a travis-ci hook anyway so whatever.
2.
3. Sounds good.  https://www.algolia.com/ provide the great search functionality on reactjs.com etc. for free.  Part of what I'm also thinking here is that a general standard feels like it's evolving (driven by FB's fairly large scale) in front end JS for that structure of marketing home page, Docs link, a few other key links in the header.  I'd like to mimic it.
4. Yes, setting in docklets would be fine.
5. We'd want something way more customised than this. We'd need to embed our own JS in the page to power it.  We want to show an editor (with support for syntax highlighting) for people to edit pug samples and another box for them to see HTML output.  We'll also want options for pretty/normal modes.  If we host on our own domain there shouldn't be any security concerns letting us embed our own JS snippets.
6. I don't think this is something to do via JSDoc but rather by extracting all the strings and letting people translate them.  The key is detecting when the english versions have been updated more recently than the translated versions and showing the english versions instead until translations catch up.  This is a big job, but would be of huge value to solve.

PR #2357, opened 2016-05-04
Fix broken Scala link

cassidoo:
The Fusesource link is dead. :)

This is the new one:
https://scalate.github.io/scalate/documentation/scaml-reference.html


TimothyGu:
Thanks!

PR #2363, opened 2016-05-07
parse inline tags as block

dutchenkoOleg:
is it possible to add an option that can be controlled manually, as "pretty" option, to be able to render inline elements as block?

my suggestion:
```
{
  "inlineAsBlock": true | false | undefined
}
```

for example:
"inlineAsBlock": false or undefined --- will not affect and render like this ▼
```
<ul>
	<li><a href="same.html">Lorem ipsum dolor.</a> <strong>maxime, dolores.</<strong></li>
</ul>
```

and "inlineAsBlock": true --- will render like this ▼
```
<ul>
	<li>
		<a href="same.html">Lorem ipsum dolor.</a>
		<strong>maxime, dolores.</<strong>
	</li>
</ul>
```

changes are also needed in
[pug-parser/index.js](https://github.com/pugjs/pug-parser/blob/master/index.js)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -62,7 +62,7 @@ function compileBody(str, options){
         return token;
       });
       tokens = stripComments(tokens, { filename: filename });
-      return parse(tokens, filename);
+      return parse(tokens, filename, undefined, options.inlineAsBlock);
     },
     read: function (filename) {
       dependencies.push(filename);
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
No, we're not making any modifications to pretty mode.  I intend to deprecate it in v3.0.0 of pug.  It's only use case is the pug documentation.  Other than that it should never be used since you don't need pretty printed output.

MattTreichelYeah:
Interesting, so what would solutions be if you planned to output HTML that you planned for further stages of human modifications to be made to?

Issue #2336, opened 2016-04-26
`.` before the last `/` should not count when checking whether a file path has an extension


artursudnik:
when extending template from parent directory, for example:
`extends ../layout `
I get error until I add ".pug" at the end:

`ENOENT: no such file or directory, open 'd:\MyCode\Nodejs\myApp\views\layout'`

ForbesLindesay:
Thanks for raising this.  It's because the file path starts with `.`.  I think we should change the code to ignore any `.` that's in the directory part of the filename.

michaelshopsin:
I can confirm this error, I have to use alpha3 instead of alpha6 until the problem is solved.

artursudnik:
I deal with it by just adding extension to template name.

TimothyGu:
@michaelshopsin, is it reproducible on Linux or do I have to use Windows?

TimothyGu:
Sorry, stupid question. The code fixing the issue has been committed but not released yet. Can you try `pug@2.0.0-alpha7`?

michaelshopsin:
@TimothyGu I can reproduce the problem on MacOS X and Linux (Redhat), I haven't tried windows. Given that `../include` is a standard construct that Jade supported it should work. I'll try `pug@2.0.0-alpha7` and see if the problem is fixed.

TimothyGu:
Okay, thanks. I was just wondering, since the original issue was filed for Windows and I don't have a Windows box near me.

michaelshopsin:
@TimothyGu no the template is still broken with alpha7, the specific template is:

    extends ../layout

    block content
       h1= title
       ...

The layout.pug file is at the top level, just like in the JADE tutorial. Under `pug@2.0.0-alpha3` the exact same code works. I call the template from Node.js with `res.render('sidecar/upload', {title: 'Sidecar Upload', data: id});` where `id` is a String.

ForbesLindesay:
This seems very surprising.  What version of node.js are you using?  Could you open a pull request with a failing test case?

You can add your test case to https://github.com/pugjs/pug/tree/master/test/cases (call it `regression.2336.pug`) and put your layout file in `auxiliary`.

michaelshopsin:
@ForbesLindesay I had the same problems with pug@2.0.0-alpha7 on Node.js 4.2.0, 6.0.0, and 6.1.0 which lead me to conclude that the problem is in pug and not Node.js. I'll put the test case into as regression.2336.pug.

[...]

PR #2375, opened 2016-05-11
Regression test for extends

michaelshopsin:
Test that shows that extends ../layout fails to include the file in
pug@2.0.0-alpha7 which worked in pug@2.0.0-alpha3 see
https://github.com/pugjs/pug/issues/2336

### DIFF
--- a/test/cases/auxiliary/layout
+++ b/test/cases/auxiliary/layout
@@ -1,0 +1,13 @@
+doctype html
+html
+  head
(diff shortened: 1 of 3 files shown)


ForbesLindesay:
Thanks for adding these.  This is working as intended.  pug defaults to adding the `.pug` extension so it includes the `.pug` file rather than the `layout` file you added which doesn't have any extension.

PR #2381, opened 2016-05-14
Add plugins API

ForbesLindesay:
I forgot we already had #2171 we should figure out the best bits of
each approach.


TimothyGu:
Looks good to me. I'll try to adapt the plugins I wrote for my version of the API and see if any problem arises.

PR #2387, opened 2016-05-17
+ 2.0 alpha history note

tomByrer:
gulp-pug 3.x with pug 2.0.0-alpha is on npm, & finding info about pug's changes takes too much effort.  Since the code updates are on this branch, so should the doc updates IMHO.

Not meant to be permanent; just to help folks for the next few weeks/months while you ramp up docs.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+2.0.0-alpha / 2016
+
+  * Better comment handling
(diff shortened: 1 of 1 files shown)


TimothyGu:
There are a lot more changes than that, so I'd rather we write it in one go than merging this PR then incrementally add them.

tomByrer:
I understand, but please update it soon since the missing changes documentation is frustrating.

PR #2488, opened 2016-08-15
Remove docs

TimothyGu:
Documentation has been moved to [pug-en](https://github.com/pugjs/pug-en), and the website [pug-www](https://github.com/pugjs/pug-www). There is now no need to keep these files in-tree.

Also removed is the client-compilation example, which is pointless as using Pug in the browser is exactly the same experience as using it in Node.js (except the `compileFile()` etc. functions, quite obviously).

The Chinese translation of the Pug syntax may be migrated to a pug-zh-CN repo, although I am more inclined to use the country-neutral pug-zh-Hans. Volunteers are of course welcome (I do speak Chinese but my time is limited).


ForbesLindesay:
We might want to wait until we have a home somewhere for the Chinese translation before merging this, even if the translation has not yet been completed.

My suggestion would be a `pug-zh` repo, in the same way as we don't have `pug-en_GB` or `pug-en_US`.  I think it is un-realistic to expect us ever to do country/dialect specific translations.

TimothyGu:
> We might want to wait until we have a home somewhere for the Chinese translation before merging this, even if the translation has not yet been completed.

That sounds fine.

> I think it is un-realistic to expect us ever to do country/dialect specific translations.

If only dialects and regional usage is concerned, probably not. For example, for Spanish you do get preferences of a specific word depending on the region, but chances are the other party will know what you are saying no matter what. American English and British English are even closer in that regard: "elevator" vs "lift"? "authorise" vs "authorize"? Come on, anyone can figure those out.

But when the writing system differ, I would say that warrant a separate handling. Check out [this](http://i2.wp.com/www.chinasmack.com/wp-content/uploads/2009/03/traditional-vs-simplified-open-picture-fun-learn.gif) for example (randomly pulled from Google). Granted, these are some extreme cases, but there are lots of times when I, as a person growing up learning simplified Chinese, want to scratch my head at a traditional character and say, _what the heck is it saying?_ (in Chinese of course). Plus, with the current political situation and whatnot I don't want to favor one system over the other.

If I come off being a bit strongly opinionated on this, sorry about that, but it's a bit frustrating to see when people not familiar with other languages making sweeping generalizations about them.

ForbesLindesay:
Fair enough.  I'm happy to defer to your judgment for Chinese and add variant suffixes where it's appropriate.

TimothyGu:
After thinking about it a bit more, and seeing both Express and Node.js are using `zh-cn`, I've decided to put the old Readme file in [pug-zh-cn](https://github.com/pugjs/pug-zh-cn).

PR #2497, opened 2016-08-22
Fix doesn't support multiple dots in file name

pixcai:
Fix Issue:

> Doesn't support multiple dots in file name while extending

See [pug-loader/issues#59](https://github.com/pugjs/pug-loader/issues/59)


TimothyGu:
https://github.com/pugjs/pug-loader/issues/59#issuecomment-241344521

> This is by design. We are also planning on requiring `.pug` for all extends/includes in the future.

PR #2498, opened 2016-08-22
support multiple dots in file name while extending

pixcai:
Support multiple dots in file name while extending


ForbesLindesay:
This is a duplicate of #2497.  It is an intentional breaking change to make include and extends consistent.

PR #2506, opened 2016-08-29
Update index.js

rottenberry:
added "var" keyword for str variable


lib/index.js
```diff
@@ -232,7 +232,7 @@ function handleTemplateCache (options, str) {
 exports.compile = function(str, options){
   var options = options || {}
 
-  str = String(str);
+  var str = String(str);
```
TimothyGu: `str` is already declared via the function parameter

Issue #2537, opened 2016-09-13
Missing ; at the end of inline CSS style


meeyuuu:
I'm having a problem creating newsletters using Pug. When the .pug files are compiled to HTML, there's always a missing ";" at the end of each inline CSS. This causes the design to break, especially on HTML newsletters for Microsoft Outlook, because of the syntax error.

Sample:

index.pug
`h1(style="color: red;") Hey!`

Result:

index.html
`<h1 style="color: red">Hey!</h1>`

ForbesLindesay:
This feels like a bug to me.  I think it was done as  a small performance optimisation, but if it breaks rendering in Outlook it's definitely not worth it.  You could fix it with a pull request to https://github.com/pugjs/pug-runtime/blob/master/index.js#L100-L117

PR #2605, opened 2016-11-01
Fix tests - pug-runtime now adds a trailing ; to inline styles #2537

jeromew:
This PR on pug tests is necessary to be compatible with the head of pug-runtime.

cf https://github.com/pugjs/pug-runtime/commit/8647d82f6faf41d7752056aef538b88c791ec29c


jeromew:
Note: these tests will pass once a new version of pug-runtime is published and once pug uses this new version

PR #2620, opened 2016-11-18
Fixed register.js to replace "\" with "/" for Windows.

Aldlevine:
Since require.resolve returns back-slashed directories in Windows, when it gets added as a string within a string (like it does in register.js) those back-slahes become escapes. Since node is perfectly happy forward-slashed directories in Windows, I just converted them.

### DIFF
--- a/register.js
+++ b/register.js
@@ -1,5 +1,5 @@
 var pug = require('./');
-var resolvedPug = require.resolve('./');
+var resolvedPug = require.resolve('./').replace(/\\/g, '/');
 
 function compileTemplate(module, filename) {
   var template = pug.compileFileClient(filename, {inlineRuntimeFunctions: false});
--- a/register.js
+++ b/register.js
@@ -1,9 +1,9 @@
 var pug = require('./');
-var resolvedPug = require.resolve('./').replace(/\\/g, '/');
+var resolvedPug = JSON.stringify(require.resolve('./'));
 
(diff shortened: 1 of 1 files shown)


TimothyGu:
Does Node not work with backslashes on Windows?

Aldlevine:
Node does work with backslashes, but the issue is because the string with backslashes is inserted into another string that then gets evaluated so the backslashes become escapes. For example, on windows node will resolve a path to 
```javascript
'C:\\project\\node_modules\\pug\\lib\\index.js'
```
 but when that's appended as a string within another string (which is what happens here), it becomes
```javascript
var pug = require('C:\project\node_modules\pub\index.js').runtime;
```
You can see that, at best the \\'s would disappear, at worst they are valid escape codes (such as the `\n` in the above example).

This pull request turns the \\'s into /'s so, it would output the below, which is perfectly valid in Windows. 
```javascript
var pug = require('C:/project/node_modules/pub/index.js').runtime;
```

TimothyGu:
@Aldlevine, ah I see the problem now. Would `JSON.stringify` fix it as well?

TimothyGu:
Thanks!

Aldlevine:
Awesome! Thanks for merging! Just for safety's sake I tested the change on both Linux and Windows and it's working flawlessly for me.

PR #2630, opened 2016-12-02
Adds naming for detection of template engine

designfrontier:
Since there is not a `constructor` and thus a `constructor.name`
for detection this provides a name that can be used for the same
purpose. This is useful for application servers that allow the
template engine to be passed in.

I am working on PRs for other template engines for this as well.
It should also be noted that Handlebars, lodash, and underscore
are all detectable through this method already.

https://github.com/olado/doT/pull/216 is a PR for this in doT
ejs: https://github.com/mde/ejs/pull/220

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -23,6 +23,12 @@ var runtime = require('pug-runtime');
 var runtimeWrap = require('pug-runtime/wrap');
 
 /**
+ * Name for detection
+ */
+ 
+exports.name = 'Pug';
+
+/**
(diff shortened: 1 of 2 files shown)


designfrontier:
as a further point of information ejs merged their pull request for this yesterday evening and it will be going out in their next release

hemanth:
Makes sense, thank you.

designfrontier:
Sweet! Thanks!

PR #2635, opened 2016-12-12
Fixes: JSDoc optionals and syntax.

sakunyo:
This Pull Request is fixed some JSDoc @param optinal and syntaxes.

sorry fix commit comment typo

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -230,7 +230,7 @@ function handleTemplateCache (options, str) {
  *   - `filename` used to improve errors when `compileDebug` is not `false` and to resolve imports/extends
  *
  * @param {String} str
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -230,7 +230,7 @@ function handleTemplateCache (options, str) {
  *   - `filename` used to improve errors when `compileDebug` is not `false` and to resolve imports/extends
  *
  * @param {String} str
- * @param {Options} options
+ * @param {Object=} options
```
ForbesLindesay: Shouldn't it be:

```
@param {Object} [options={}]
```

(and ditto for the others?

Issue #2675, opened 2017-01-22
Convert to a monorepo


ForbesLindesay:
I think a monorepo (but still with separate npm packages) will make development on pug much easier & faster.  I propose that we move the following packages into a single monorepo:

- pug
- pug-lexer
- pug-parser
- pug-load
- pug-linker
- pug-code-gen
- pug-filters
- pug-runtime
- pug-walk
- pug-strip-comments
- pug-ast-spec
- pug-error
- pug-attrs

I propose that we keep the following packages separate:

- constantinople
- doctypes
- doc-down
- is-expression
- is-expression-babylon
- pug-www
- pug-en, pug-zh-cn etc.
- void-elements
- babel-plugin-transform-react-pug
- gulp-pug
- pug-cli
- pug-lint
- with
- then-pug (still currently published as then-jade but needs renaming)
- token-stream
- babel-plugin-transform-with
- babylon-walk
- all pug plugins
- pug-lint-config-migration-v2
- pug-logo
- react-pug (still currently published as react-jade but needs renaming)
- pug-plugin-dynamic-include
- pug-source-gen
- pug-plugin-debug-line
- brjade
- js-stringify

I think some of these packages will probably move into the monorepo over time if it proves successful.  We will continue to follow semver individually in each package we release.

ForbesLindesay:
Each of the repositories now in the monorepo needs to have issues coppied & pull requests coppied over to the monorepo, have issues disabled, and be replaced with a simple README explaining where the project has moved to:

- [ ] pug-lexer
- [ ] pug-parser
- [ ] pug-load
- [ ] pug-linker
- [ ] pug-code-gen
- [ ] pug-filters
- [ ] pug-runtime
- [x] pug-walk
- [ ] pug-strip-comments
- [ ] pug-error
- [ ] pug-attrs

Look at pug-walk for an example of what needs doing.

jeromew:
I migrated https://github.com/pugjs/pug-code-gen but I don't seem to have rights to disable new issues on this repo.

ForbesLindesay:
Thanks, I've disabled them.

OwnageIsMagic:
https://github.com/blog/2460-archiving-repositories

PR #2676, opened 2017-01-23
Monorepo

ForbesLindesay:
As described in #2675.  This combines many of pug’s separate packages into a single monorepo and replaces the various test runners with jest.  Although a few years ago, jest was pretty terrible, it’s recently become awesome.  It has out of the box support for:

 - **snapshots** - these have great diff views, and an easy mechanism for updating them when they change, letting us discard lots of our custom code for snapshotting the output of pug.
- **watch mode** - this completely changes the experience of using tests when authoring code.
- **coverage** - there is no need for awkward hacks or a separate tool, just pass `—coverage`.
- **monorepos** - because jest finds all files ending in `.test.js`, it works perfectly in a monorepo.  It saves us from needing to treat each package separately.
- **performance** - jest runs all the different files in parallel, meaning that running all the tests is pretty much as fast as just running the existing tests for only the main pug module.


packages/pug-attrs/.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: node_js
```
TimothyGu: These individual `.travis.yml`s can now be removed.

thejameskyle:
Dunno if you already saw this but lerna also has a `lerna import <path/to/repo>` command for maintaining git history while merging repos in.

ForbesLindesay:
Yes, the command was erroring on my local machine, and I decided I didn't really want all that history anyway.  The code has gone through so many big refactorings that the low level detail of all those individual commits is not very useful.

ORESoftware:
Hope this doesn't turn out to be a nightmare! mono-repo ftw!

PR #2679, opened 2017-01-23
[pug-filters] Apply options to nested filters

ForbesLindesay:
Original fix authored by @zah

This is a simple bug fix.

``` pug
script
  :cdata-js:babel(presets=['es2015'])
    const myFunc = () => `This is ES2015 in a CD${'ATA'}`;
```

When nested filters are used, the inner filter (babel) was failing to obtain its options from the `options.filterOptions.babel` table.

### DIFF
--- a/scripts/filename-serializer.js
+++ b/scripts/filename-serializer.js
@@ -1,0 +1,20 @@
+const path = require('path');
+const basedir = path.resolve(__dirname + '/..');
+
+// filename serializer that removes the basedir
+module.exports = {
+  test: function(val) {
+    return (
+      val && typeof val === 'object' && ('filename' in val) &&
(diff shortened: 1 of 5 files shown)


scripts/filename-serializer.js
```diff
@@ -0,0 +1,20 @@
+const path = require('path');
+const basedir = path.resolve(__dirname + '/..');
+
+// filename serizlier that removes the basedir
```
TimothyGu: typo

packages/pug-filters/test/per-fitler-options-applied-to-nested-filters.test.js
```diff
@@ -0,0 +1,28 @@
+const lex = require('pug-lexer');
```
TimothyGu: typo in filename

zah:
Thanks guys, I've been using this for a while in a big project without issues.

PR #2680, opened 2017-01-23
Updated package.json adding license as well as the repository

mfainshtein4:
noticed that the package.json was missing some information so I figured id populate a few of the missing fields.


package.json
```diff
@@ -6,6 +6,10 @@
     "jest": "^18.1.0",
     "lerna": "2.0.0-beta.32"
   },
+  "repository":{
+	"type": "git"
```
ForbesLindesay: Missing a comma here.

package.json
```diff
@@ -6,6 +6,10 @@
     "jest": "^18.1.0",
     "lerna": "2.0.0-beta.32"
   },
+  "repository": {
+	"type": "git",
+	"url": "https://github.com/pugjs/pug.git"
```
ForbesLindesay: The indentation looks wrong here.

package.json
```diff
@@ -6,6 +6,10 @@
     "jest": "^18.1.0",
     "lerna": "2.0.0-beta.32"
   },
+  "repository": {
+	  "type": "git",
+	  "url": "https://github.com/pugjs/pug.git"
```
ForbesLindesay: This still needs to be indented by another space.

PR #2681, opened 2017-01-24
[pug-filters] Add a filterAliases option

### LINKED ISSUE #2678: Support filter aliases
It would be good if there was an option you could pass in to say that `:markdown` should use `jstransformer-markdown-it`, making it easier to swap markdown implementation later.

ForbesLindesay:
[closes #2678]

This provides a few things in one neat package:

1. It provides a cleaner upgrade path from old versions.  Instead of updating everywhere you reference `:markdown` in your source code, just add `filterAliases: {markdown: ‘markdown-it’}`.
2. It makes it really easy to swap out different implementations of essentially the same filter (e.g. all the different markdown parsers).
3. Because `filterOptions` are processed **before** aliases, it lets you use aliases to define presets.  e.g. you could make `minify` and `beautify` both aliases of `uglify-js` but then apply different `fitlerOptions` to each alias.


TimothyGu:
Is this better implemented as a plugin?

ForbesLindesay:
I don't think so.  In order to support the definition of presets it needs to be part of the existing filters module.

There may be merit to making filters into a plugin, that is enabled by default, but I don't think we should try to extract this specific feature into a plugin.

PR #2685, opened 2017-01-25
[pug-linker] Duplicate block

ForbesLindesay:
This allows block names to be duplicated.


ForbesLindesay:
I'm reasonably happy with this code.  Long term we should look at refactoring the linker and making it simpler, if we can.  In the mean time, this regression has existed for too long so I'm going to merge it now.

PR #2687, opened 2017-01-25
[pug-linker] Add regression tests for #1261

ForbesLindesay:
This adds a test demonstrating that some of the issues reported in #1261 are still present.  The snapshot will need updating once the issue is fixed.


ForbesLindesay:
I think the snapshots I've pushed now represent the behaviour the user expected to see.  The only trouble is, I'm not really sure the output was "wrong" to start with.  The issue is that `append` and `prepend` only really make sense at the top level of a file.  By supporting them in nested locations, we end up creating a block as well as appending to the existing block.  The simplest solution would be to disallow append/prepend except at the top level.  I'm not confident that won't break some people's use cases though.

PR #2691, opened 2017-01-25
Remove ES6 syntax from pug-linker

davidminor:
Change 'fat arrow' function into regular function syntax. This will
make it simpler for codebases to depend on pug if they are using older
versions of node.


ForbesLindesay:
@TimothyGu Do you want to run the release, or should I?

TimothyGu:
@ForbesLindesay, go ahead and do it. Kinda busy recently, and I'm not sure I'm confident with publishing with monorepo.

ForbesLindesay:
ok, I'm doing it now.  It's actually really easy: just run `npm run release` and answer the questions you get asked by lerna.

PR #2693, opened 2017-01-26
Fix unit tests on Windows machines

jbsulli:
I was not able to run the unit tests on my Windows machine. Turns out it was an easy fix.

### DIFF
--- a/packages/pug/test/run-syntax-errors.test.js
+++ b/packages/pug/test/run-syntax-errors.test.js
@@ -9,7 +9,7 @@ describe('certain syntax is not allowed and will throw a compile time error', fu
   anti.forEach(function(test){
     var name = test.replace(/[-.]/g, ' ');
     it(name, function(){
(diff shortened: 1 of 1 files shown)


jbsulli:
As a side note, I've always used Node's `path.join()` to keep paths logical across operating systems:

```javascript
var path = require('path');

var p = path.join(__dirname, 'anti-cases', test + '.pug');
// C:\github\pug\packages\pug\test\anti-cases\unclosed-interpolated-tag.pug

// vs

var path = __dirname + '/anti-cases/' + test + '.pug';
// C:\github\pug\packages\pug\test/anti-cases/unclosed-interpolated-tag.pug


```

ForbesLindesay:
I normally just work on the grounds that node already handles `/` style file paths fine on windows.  The jest serializer also just handles the file paths for us in snapshots: https://github.com/pugjs/pug/blob/master/scripts/filename-serializer.js

PR #2712, opened 2017-02-08
[pug-lexer][optimisation] Only call assertExpression when necessary

ForbesLindesay:


ForbesLindesay:
I haven't benchmarked this.  It would be good if someone felt like having a go at that.  I would expect it to be a significant performance improvement though.

Issue #2717, opened 2017-02-10
Support for coffee-script as code


aliksend:
Hi

Is it difficult to allow using coffeescript as [code](https://pugjs.org/language/code.html).
I mean thing like that:
```
div
   -:coffee-script
      a = b if c is d
```

It is means: process coffee-script code using filter, get js and run it while processing template as common js code:
```
div
   -
      if (c == d) {
         a = b;
      }
```

aliksend:
To detect it:
packages/pug-parser/index.js:577
```
var text = '';
    var preFilter = '';
    if (body.type === 'filter') {
      var filter = this.advance();
      preFilter = filter.val;
      // re-peek body
      body = this.peek();
    }
    if (body.type === 'start-pipeless-text') {
```

packages/pug-parser/index.js:607
```
    return {
      type: 'Code',
      preFilter: preFilter,
      val: text,
      buffer: false,
```

Remains to find where the code is executed.
May be I'll create pull request later

PR #2718, opened 2017-02-10
Using filters with code

aliksend:
Allowing using filters with code in templates
#2717

### DIFF
--- a/packages/pug-filters/lib/handle-filters.js
+++ b/packages/pug-filters/lib/handle-filters.js
@@ -32,6 +32,9 @@ function handleFilters(ast, filters, options, filterAliases) {
       });
       node.filters = undefined;
       node.file = undefined;
+    } else if (node.type === 'Code' && node.preFilter) {
+      var attrs = { filename: node.filename };
+      node.val = filterWithFallback(node, node.val, attrs);
     }
 
     function filterWithFallback(filter, text, attrs, funcName) {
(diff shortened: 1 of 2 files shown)


aliksend:
There is no tests for this feature yet

ForbesLindesay:
I'm a little concerned about this bloating the language.  It's yet another thing to document and for people to learn when new to the language.  It is neat though, I'm just not convinced many people are so attached to Coffee Script that they would want to add `:coffee-script` before every statement.

If we were going to do it, we'd need to support it absolutely everywhere we support JavaScript:

 - interpolation
 - inline JavaScript
 - attribute values
 - buffered code (escaped and unescaped)
 - unbuffered code
 - mixin arguments
 - ...

I'm not sure that's really justified.

ForbesLindesay:
P.S. this may be something that could work well as a plugin.

aliksend:
I think that it is not bloating the language. I propose to look at that as small feature, not a part of language. And it is not about coffee, it is about combining filters (any filter) and blocks of code.

I think that it isn't necessarily to support this feature everywhere. In most parts js will be enough (and it will look exactly as coffee in 90% cases in short lines of code). But in big blocks of code this improvement will allow to use coffee instead of js. That developer, who want to use it, should understand that code will be compiled to js and then executed.
Documentation (on code section) can be like "also you can use filters with blocks of code" and small example.

Plugin is a good idea, but it will be a lot of work.

PR #2761, opened 2017-03-17
Add compileFileModule method

guumo:
Add compileFileModule method. This method is very useful for compiling templates to modules compatible with node js in order to make a require. In this way, we can save the content in a file for later use as a standalone module or asynchronous cache.

### DIFF
--- a/packages/pug/lib/index.js
+++ b/packages/pug/lib/index.js
@@ -276,6 +276,7 @@ exports.compile = function(str, options){
  *     the compiled template for better error messages.
  *   - `filename` used to improve errors when `compileDebug` is not `true` and to resolve imports/extends
  *   - `name` the name of the resulting function (defaults to "template")
+ *   - `module` when it is explicitly `true`, the source code include export module syntax
  *
  * @param {String} str
  * @param {Options} options
@@ -303,6 +304,7 @@ exports.compileClientWithDependenciesTracked = function(str, options){
     filterOptions: options.filterOptions,
     filterAliases: options.filterAliases,
     plugins: options.plugins,
+    module: options.module
   });
 
   return {body: parsed.body, dependencies: parsed.dependencies};
--- a/packages/pug/lib/index.js
+++ b/packages/pug/lib/index.js
@@ -181,8 +181,7 @@ function compileBody(str, options){
     globals: options.globals,
     self: options.self,
     includeSources: options.includeSources ? debug_sources : false,
-    templateName: options.templateName,
-    module: options.module
+    templateName: options.templateName
   });
   js = applyPlugins(js, options, plugins, 'postCodeGen');
 
@@ -304,11 +303,19 @@ exports.compileClientWithDependenciesTracked = function(str, options){
     filters: options.filters,
     filterOptions: options.filterOptions,
     filterAliases: options.filterAliases,
-    plugins: options.plugins,
-    module: options.module
+    plugins: options.plugins
   });
 
-  return {body: parsed.body, dependencies: parsed.dependencies};
+  var body = parsed.body;
+
+  if(options.module) {
+    if(!options.inlineRuntimeFunctions) {
+      body = 'var pug = require("pug-runtime");' + body;
+    }
+    body += ' module.exports = ' + (options.templateName || 'template') + ';';
+  }
+
+  return {body: body, dependencies: parsed.dependencies};
 };
 
 /**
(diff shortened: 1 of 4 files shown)


ForbesLindesay:
Can we make this into an option in the `compileClientWithDependenciesTracked` method.  That way we won't need separate `compileModule`, `compileFileModule` etc. methods.  It will also make it easier to add a `--module` option to the CLI (and gulp, grunt etc.)

It should also detect the value of `inlineRuntimeFunctions` and only add the `require('pug-runtime')` call if `inlineRuntimeFunctions` is `false`.

guumo:
I agree! 👍 👍 👍   Shall I try to implement the changes you mentioned and make a new pull request with it or will you make it ? Thanks!!!

ForbesLindesay:
You can either make a new pull request with the changes, or update this one.  If you update this one, please comment on it to let me know it needs reviewing again.

guumo:
Hi, @ForbesLindesay! I just uploaded the changes with a new test case, I don't know if it's fine. I'll be waiting for your feedback, thanks!! :)

packages/pug/lib/index.js
```diff
@@ -303,6 +305,7 @@ exports.compileClientWithDependenciesTracked = function(str, options){
     filterOptions: options.filterOptions,
     filterAliases: options.filterAliases,
     plugins: options.plugins,
+    module: options.module
   });
 
   return {body: parsed.body, dependencies: parsed.dependencies};
```
ForbesLindesay: Please make your changes **here**.  i.e. add some code like:

```js
var body = parsed.body;
if (options.body) {
  // wrap body in module here
}
return {body: body, dependencies: parsed.dependencies};
```

That way we don't make pug-code-gen any more complex.

packages/pug/test/pug.test.js
```diff
@@ -979,12 +979,20 @@ describe('pug', function(){
...
-      var src = '.bar= self.foo'
+      var src = '.bar= self.foo';
       var fn = pug.compileClient(src, {self: true});
       fn = Function('pug', fn.toString() + '\nreturn template;')(pug.runtime);
       var actual = fn({foo: 'baz'});
       assert(actual === '<div class="bar">baz</div>');
     });
+    it('should support module syntax .pug.compileClient(str, options)', function () {
+      var src = '.bar= self.foo';
+      var fn = pug.compileClient(src, {self: true, module: true, inlineRuntimeFunctions: false});
```
ForbesLindesay: It would be good to have an example with both `inlineRuntimeFunctions: true` and `inlineRuntimeFunctions: false`.

You might also want to add `expect(fn).toMatchSnapshot();` which will make the generated code part of the test, making it easier to see if this functionality gets broken in some subtle way.

packages/pug/test/pug.test.js
```diff
@@ -979,12 +979,20 @@ describe('pug', function(){
...
+      fs.writeFileSync(__dirname + '/temp/input-compileModuleFileClient.js', fn.toString());
```
ForbesLindesay: `compileClient` should already return a string, so you shouldn't need to call `.toString()`, other tests may be out of date though.

guumo: Ok, I added toString() because I saw that in the previous test. 

 `fn = Function('pug', fn.toString() + '\nreturn template;')(pug.runtime);`

packages/pug/test/pug.test.js
guumo: Ok, I added the second test. I have doubts about how to implement this `expect(fn).toMatchSnapshot();` because in the `.compileClient()` Test I didn't see how this can be used.

packages/pug/lib/index.js
guumo: Ok, I have done the changes.

guumo: I need help here:
`templateName: options.name || 'template',`
I don't understand why `options.name` is not `options.templateName`
Thanks! :)

guumo:
I just commit the snapshots. Thanks!

ForbesLindesay:
Thanks for sticking with this through those changes. If you have the time, it would be great if you could also add documentation for this option to https://github.com/pugjs/pug-en/blob/master/src/api/reference.md

guumo:
Ok! I will work to add it to the documentation :) thanks!

ForbesLindesay:
Just published this as 2.0.0-rc.2

PR #2788, opened 2017-04-23
Omit optional HTML5 tags

jart:
See: https://www.w3.org/TR/html5/syntax.html#optional-tags

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,22 +80,17 @@ becomes
 
 
 ```html
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
-  <head>
-    <title>Pug</title>
-    <script type="text/javascript">
-      if (foo) bar(1 + 5)
-    </script>
-  </head>
-  <body>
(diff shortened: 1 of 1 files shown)


jart:
Here's a good reference implementation of the standard if you want to make your HTML generator omit the optional tags. https://github.com/bazelbuild/rules_closure/commit/6c5cbe3d51ccebe0f47ec207280533f2e80d1142

ForbesLindesay:
I think it is better if we generate all the optional tags. It makes our output easier to understand and more consistent.

jart:
It's so much harder to understand HTML if you have to read through all that superfluous opening and closing tag boilerplate.

Now that I've brought to your attention that a more elegant output format exists, and provided you with an Apache 2.0 licensed implementation you can loosely copy, will you decline to implement the more concise, succinct, and elegant solution?

ForbesLindesay:
Yes, I will.

PR #2792, opened 2017-04-26
Remove named blocks from included templates if they extend

### LINKED ISSUE #2436: same named block in nesting extends file
page.pug

```
extends layout.pug

block foo
    h1 page text

block widgets
    include widget.pug
```

widget.pug

```
extends widget_base.pug

block foo
    h3 widget text
```

the block `foo` in `page.pug` will be `<h3>widget text</h3>`.

pug 2.0.0-alpha3

is this a bug?

ForbesLindesay:
If an included template extends some other template, the named blocks should be removed once it has been linked.  This will stop those named blocks interfering with named blocks in the main template chain.

[fixes #2436]
[closes #2697]

### DIFF
--- a/packages/pug-linker/index.js
+++ b/packages/pug-linker/index.js
@@ -55,6 +55,7 @@ function link(ast) {
       parent.declaredBlocks[name] = ast.declaredBlocks[name];
     });
     parent.nodes = mixins.concat(parent.nodes);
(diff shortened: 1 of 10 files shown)


ForbesLindesay:
Sorry, I've messed up the rebasing here.  Only the first and last commit are actually relevant.

PR #2812, opened 2017-06-02
[pug-parser] Remove excess blocks

ForbesLindesay:
This has a tiny effect on the output of the pretty printed mode, but
mostly it just simplifies rendering for third things like react-pug.


Bratanov:
@ForbesLindesay Hey, I'm using pug/pug-parser as a dependency of a module (kue) in my project and using `const` here breaks the node v0.12 (non-es6) support. Version 2.0.0-beta9 (and maybe 10) worked fine for me.
Is it intended for Pug to be ES6+ only, or was this unintentional?

ForbesLindesay:
Pug supports node v4.0 and later.

PR #2853, opened 2017-08-15
Keep newline between text and interpolated-code

akihikodaki:
Here, the following example will have a newline between `foo` and `bar`.
```pug
| foo
| bar
```

However, the following will not.
```pug
| foo
| #{bar}
```

This pull request fixes the behavior.

# Question
Should it also keep newline between text and code? For example,
```pug
| foo
= bar
```

This still won't have a newline.

### DIFF
--- a/packages/pug-parser/index.js
+++ b/packages/pug-parser/index.js
@@ -304,7 +304,8 @@ Parser.prototype = {
           case 'newline':
             if (!options || !options.block) break loop;
             var tok = this.advance();
-            if (this.peek().type === 'text') {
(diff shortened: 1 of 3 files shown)


ForbesLindesay:
Thanks, the behaviour for:

```
| foo
= bar
```

is correct.  I think:

```
| foo
| #[bar]
```

should probably preserve the newline though, so if you could check that and submit a pull request/fix that would be awesome.

akihikodaki:
I have confirmed tag interpolation preserves preceding newlines. It does not need any fix.

PR #2912, opened 2017-11-23
Fix #2911

### LINKED ISSUE #2911: Pug + RJS compile issue
**Pug Version:**  2.0.0-rc.4

**Node Version:**  9.2.0

## Input JavaScript Values

```js
pug.compileClient(str, { filename: 'filter.pug' });
```

## Input Pug

```pug
.filter.filter-container.hl-filter
	.input-container.prop-val-container.form-element.mature.filled.placeholder-dark.pointer-parent
		i.down-arrow.icon-arrow-down
		span.rupee-symbol.icon-rupee
		input.input.prop-val(type = 'select' placeholder = "Property Value" required = true pattern = "[0-9,]{1,12}")
		.select.select-outer.small

.filter.filter-container.hl-filter
	.input-container.prop-type-container.form-element.inline-el.filled.placeholder-dark.pointer-parent
		i.down-arrow.icon-arrow-down
		input( class = 'input hl-service-select' type = 'select', placeholder = 'Property Type', required = true, readonly=true)
		.select.select-outer.small

.filter.filter-container.hl-city-filter
	.input-container.city-select-container.form-element.inline-el.filled.pointer-parent
		i.icon.clear-city-text.icon-close-circle
		.pointer.pointer-fake
		.pointer
		input( class = 'input hl-city-select' type = 'select', placeholder = 'Property City', required = true)
		.select.select-outer.small

button.btn.primary.view-offers-btn.btn-search View Offers
```

After compilation, I get [this](https://gist.github.com/bogas04/7760c576584ea78517b55aeb17c40864) javascript file. This file seems to have a [syntax issue](http://esprima.org/demo/validate.html), which gets caught by mine RJS Optimizer.

From what I understand, the return statement has a function `pug_rethrow`, immediately followed by `function template()`. The compiler is complaining about missing `;` or `\n` after end of `pug_rethrow`.

```js
function () {
  return function pug_rethrow(){/* some awesome stuff */}function template(){/* more awesome stuff */}
}
// -----------------------------------------------------^
// Unexpected token function
// Adding `;` or `\n` fixes it
```
## Additional Comments

I'm not sure if RJS is causing this issue or pug. Maybe pug outputs `\n` but RJS minifies it incorrectly? Anyway it'll be nice if pug can produce correct syntax by using `;` over `\n` to avoid this scenario completely. 

I guess the culprit is here:

https://github.com/pugjs/pug/blob/master/packages/pug-code-gen/index.js#L147-L151

```diff
        ');' +
        '}';
    }
-    return buildRuntime(this.runtimeFunctionsUsed) + 'function ' + (this.options.templateName || 'template') + '(locals) {var pug_html = "", pug_mixins = {}, pug_interp;' + js + ';return pug_html;}';
+    return buildRuntime(this.runtimeFunctionsUsed) + ';function ' + (this.options.templateName || 'template') + '(locals) {var pug_html = "", pug_mixins = {}, pug_interp;' + js + ';return pug_html;}';

},

````

### PR DESCRIPTION
**Pull request by bogas04**
Fix for #2911 

Minifications can lead to invalid syntax. To be safe, adding this extra semi-colon will ensure that `buildRuntime(this.runttimeFunctionsUsed)` wouldn't cause `function template()` to be at wrong place, when generated by [code-gen](https://github.com/bogas04/pug/blob/f8d86d8670faa5b69ba286c835f8e9c231bdeb4f/packages/pug-code-gen/index.js#L147-L150), such as [this](https://gist.github.com/bogas04/7760c576584ea78517b55aeb17c40864) one.

```js
function () {
  return function pug_rethrow(){/* some awesome stuff */}function template(){/* more awesome stuff */}
}
// -----------------------------------------------------^
// Unexpected token function
// Adding `;` or `\n` fixes it
```

### DIFF
--- a/packages/pug-code-gen/index.js
+++ b/packages/pug-code-gen/index.js
@@ -147,7 +147,7 @@ Compiler.prototype = {
         ');' +
         '}';
     }
-    return buildRuntime(this.runtimeFunctionsUsed) + 'function ' + (this.options.templateName || 'template') + '(locals) {var pug_html = "", pug_mixins = {}, pug_interp;' + js + ';return pug_html;}';
+    return buildRuntime(this.runtimeFunctionsUsed) + ';function ' + (this.options.templateName || 'template') + '(locals) {var pug_html = "", pug_mixins = {}, pug_interp;' + js + ';return pug_html;}';
   },
 
   /**


ForbesLindesay:
The lack of newlines is not the issue.   Adding them (or semicolons) would not make any difference as you would still be returning the wrong function. Pug is not adding the `define(...)` wrapper and it is that part that is wrong.  Instead of `return RESULT_OF_COMPILE_CLIENT` you need:

```
RESULT_OF_COMPILE_CLIENT
return template;
```

inside the define function.

PR #2955, opened 2018-02-11
Ignore *lock file

ezhlobo:
After installing dependencies (npm@5.5.1) it automatically adds 'package-lock.json' file to each sub-project. I guess that we don't have lockfiles intentionally, so I asked git to ignore such files.

If that is not what we want, let's track lockfiles in git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ npm-debug.log
 **/test/temp
 .release.json
 lerna-debug.log
+package-lock.json


ForbesLindesay:
The project should work better if you use yarn instead of npm, since yarn's support for monorepos is much better.

ezhlobo:
@ForbesLindesay yarn.lock is not in .gitignore, should we add it as well?

ForbesLindesay:
Yes

PR #2956, opened 2018-02-11
Don't ask escaping for boolean attributes at the beginning of attributes-list

ezhlobo:
Each attribute without value has `true` value by default. It should not be 'escaped' because it happens internally. But when we place attribute without a value at the beginning, we have a bug when `mustEscape` is set to `true`.

This PR prevents this behavior.

## Example

**Current**

```pug
div(first, second, third)

//- first  -> mustEscape: true
//- second -> mustEscape: false
//- third  -> mustEscape: false
```

**Next**
```pug
div(first, second, third)

//- first  -> mustEscape: false
//- second -> mustEscape: false
//- third  -> mustEscape: false
```

### DIFF
--- a/packages/pug-lexer/index.js
+++ b/packages/pug-lexer/index.js
@@ -1019,7 +1019,7 @@ Lexer.prototype = {
       var whitespaceRe = /[ \n\t]/;
       var quoteRe = /['"]/;
 
-      var escapedAttr = true
+      var escapedAttr = false
       var key = '';
       var val = '';
       var state = characterParser.defaultState();
--- a/packages/pug-lexer/test/__snapshots__/index.test.js.snap
+++ b/packages/pug-lexer/test/__snapshots__/index.test.js.snap
@@ -634,7 +634,7 @@ Array [
   Object {
     "col": 3,
     "line": 3,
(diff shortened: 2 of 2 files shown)


ezhlobo:
Maybe I understood this flag wrongly, and instead of `false` it should be `true`. Could you give me feedback, please?

jbsulli:
Yeah, I've observed this behavior as well. We had a discussion on it here: https://github.com/pugjs/pug-lexer/issues/76#issuecomment-274359819. The pull request that I am working on is going to make all of the attributes without a value default to `true` based on their feedback.

ezhlobo:
@jbsulli oh, thank you for letting me know. I do need this fix to improve other project so what do you think: should I wait for your changes or make this PR appropriate?

ForbesLindesay:
Thanks, I think consistency is the main thing here, and I think it's fine to mark these as unescaped.

ezhlobo:
@ForbesLindesay hey, are you planning to publish this fix?

ForbesLindesay:
I'm going to wait for #2957 as I think we might flip to marking all these attributes as "mustEscape: true".

PR #2957, opened 2018-02-12
pug-lexer `loc` with source endings

jbsulli:
* Instead of `line`/`col`, there is now a `loc` object on all tokens that matches the babel `loc` object (ex: `{start:{line:1,column:1},end:{line:1,column:10},filename:"test.pug"}`
* Also assigns `mustEscape: true` to all boolean attributes (ex: `input(disabled)`)
* Refactor `attrs` into three separate functions: `attrs`, `attributeName` and `attributeValue`
* Update `pug-parser` and `pug` to use new `loc` from lexer tokens


packages/pug-lexer/index.js
```diff
@@ -990,172 +1046,244 @@ Lexer.prototype = {
...
-        , str = this.input.substr(1, index-1);
-
+  
+  attributeName: function(str){
+    var quote = '';
+    var quoteRe = /['"]/;
+    var key = '';
+    var i;
+    
+    characterParser.defaultState();
```
ForbesLindesay: This line is a no-op. You don't need to call `defaultState` unless you are going to use the state that gets returned.

packages/pug-lexer/index.js
```diff
@@ -990,172 +1046,244 @@ Lexer.prototype = {
...
+  attributeValue: function(tok, str){
+    var quoteRe = /['"]/;
+    var val = tok.val = '';
+    var done, i, x;
+    var escapeAttr = true;
+    var state = characterParser.defaultState();
+    var col = this.colno;
+    var line = this.lineno;
+    
+    characterParser.defaultState();
```
ForbesLindesay: Again, this line is a no-op

packages/pug-lexer/index.js
```diff
@@ -990,172 +1046,244 @@ Lexer.prototype = {
...
+      }
+    }
+    
+    this.assertExpression(val);
+    
+    this.lineno = line;
+    this.colno = col;
+    
+    tok.val = val;
+    tok.mustEscape = escapeAttr;
```
ForbesLindesay: Rather than taking and mutating a token. I feel if would be clearer if this function returned `{val, mustEscape, remainingSource}`

packages/pug-lexer/index.js
```diff
@@ -990,172 +1046,244 @@ Lexer.prototype = {
...
-
-  attrs: function() {
-    if ('(' == this.input.charAt(0)) {
-      var startingLine = this.lineno;
-      this.tokens.push(this.tok('start-attributes'));
-      var index = this.bracketExpression().end
-        , str = this.input.substr(1, index-1);
-
+  
+  attributeName: function(str){
```
ForbesLindesay: This function should be called `attribute` since calling it has the effect of lexing both the name **and** the value. In the future, we might split out a separate `attributeName` function.

packages/pug-lexer/index.js
```diff
@@ -990,172 +1046,244 @@ Lexer.prototype = {
...
 
-      // Reset the line numbers based on the line started on
-      // plus the number of newline characters encountered
-      this.lineno = startingLine + (str.match(/\n/g) || []).length;
-
-      this.tokens.push(this.tok('end-attributes'));
+      tok = this.tok('end-attributes');
       this.incrementColumn(1);
+      this.tokEnd(tok);
+      this.tokens.push(tok);
```
ForbesLindesay: In most places you have done `this.tokens.push(this.tokEnd(tok));` - I think we should be consistent here.

ForbesLindesay:
I added some comments to the commit. I think this looks really good, but it's a bit tricky to review because there's so much git merging happening and it makes it hard to figure out exactly what's changed. If you could rebase everything off master so there's only a few commits. Ideally have the commit that updates all the snapshots be a separate commit from the commit with your manual changes to the source code, that would really help, but if that's too much work, then that's ok.

jbsulli:
Just updated the commits and removed the merge commits. I will update with your suggestions when I get a chance later today.

ForbesLindesay:
Ping me when you get a chance to do those updates.

jbsulli:
@ForbesLindesay, I just finished the updates you requested. 

For PR #2956, it's easy to default `mustEscape` to false for boolean attributes. Just change the value here: (packages/pug-lexer/index.js:1122).

ForbesLindesay:
Thanks 👍

PR #2970, opened 2018-03-02
Add possibility to pass custom options to plugins

SerayaEryn:
Currently only certain options are passed to `compileBody` and therefore it is not possible pass custom options to plugins.

This adds a possibly to pass custom options to plugins by adding a property named `pluginOptions` to the options that are passed to `compileOptions`.

### DIFF
--- a/packages/pug/lib/index.js
+++ b/packages/pug/lib/index.js
@@ -255,6 +255,7 @@ exports.compile = function(str, options){
     filterOptions: options.filterOptions,
     filterAliases: options.filterAliases,
     plugins: options.plugins,
+    pluginOptions: options.pluginOptions
   });
 
   var res = options.inlineRuntimeFunctions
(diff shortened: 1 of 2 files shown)


ForbesLindesay:
My intuition is that it would be preferable to encourage an API like:

```js
someOptions => plugin
```
to be exposed by the plugin module. That way pug wouldn't need to handle any of the plugin options.

ForbesLindesay:
e.g.

```js
    var pugLoadPlugin = pluginOptions => ({
      resolve: (filename, source, options) => {
        assert(pluginOptions !== undefined);
        assert(pluginOptions.test === 'something');
        return pugLoad.resolve(filename, source, options);
      }
    });
    const options = {
      plugins: [
        pugLoadPlugin({test: 'something'})
      ],
      filename: filename
    };

    pug.compile(str, options);
```

That approach should pass as is.

Issue #2547, opened 2016-09-17
runtime.js not existing.


bagbag:
Hello,

i just wanted to pug on client side and looked at [https://pugjs.org/api/reference.html](url).

There it says 

> Be sure to send the Pug runtime (node_modules/pug/runtime.js) to the client in addition to the template that you just compiled.

But i can't find that runtime.js.

cekvenich:
I do not think runtime.js is needed or means anything, just a typo.

Here is a working example:
http://github.com/cekvenich/exWebCompSass/tree/master/nlib

TimothyGu:
We haven't yet updated that set of instructions to Pug. Thanks for pointing this out.

bagbag:
@cekvenich 
Either I am dumb or that example doesn't show me, how to use pug on client side (in browser).

cekvenich:
I want to help but sometimes I am short. 
Browser only runs HTML and CSS (and .js) - that is it. For example if you Sass, you must 'transpile' for CSS. If you Pug, you must 'transpile' for html. 
The confusing way that there are many ways  transpile. Ex: PrePros. 
My email is vin (at) eml.cc if you want to screen share so I can show you.

TimothyGu:
@cekvenich, please don't spread things that you don't completely understand (and this issue tracker isn't a job ad).

@bagbag, there are two kinds of "how to use pug on client side".

1. **Rendering** Pug templates that are precompiled
2. **Compiling** Pug templates in the browser

Which one do you need?

bagbag:
My goal is to dynamically create HTML in browser, without the need of a server dealing with pug.

Precompiling wouldn't be a problem for me.

cekvenich:
@TimothyGu this issue tracker isn't a job ad

Why do you say that, Because I tried to help?

ForbesLindesay:
@TimothyGu: I'm not sure what you mean re job ad.  I think @cekvenich was just trying to help.

@bagbag The docs still haven't really been written for this.  The best way to do this is to precompile templates on the server.  You can do this using the pug cli:

```
npm install pug-cli -g
pug --client --name myTemplateFunction my-pug-file.pug
```

This will produce a JavaScript file called `my-pug-file.js` which defines a template function called `myTemplateFunction`.  You can then use it like:

```js
container.innerHtml = myTemplateFunction({someVariables: 'with values'});
```

[...]

Issue #2407, opened 2016-05-31
Client Compilation


jescalan:
Just wanted to see if anyone had a moment to quickly describe the status of `compileClient` in pug. I pulled down the alpha and messed with it a little bit, only to find that things were quite different than they are in current jade.

What I noticed so far:

- No more `runtime.js`, but there is a `pug-runtime` library. Not entirely sure if it's still needed to run a client-compiled file. It seems like it is not, from my testing so far.
- Client-compiled strings are exported along with some helpers, as a named function called `template`, which can be changed using a `templateName` option. However, this option doesn't actually work and it's always exported as `template`.
- If you prefix a client template with a `module.exports =` it no longer works with a commonjs-based system, since it exports one of the helpers instead.
- If every client compile is exported along with its helpers, this is going to be very inefficient if you are trying to put multiple templates on a page. I thought this was the reason for the runtime library?
- Instead of using characters like `<` it uses unicode escape sequences like `\u003C`. This looks very strange when written to a file. It also does not render correctly when written to a page, maybe this is a bug?

It's possible I'm doing something terribly wrong, but none of the docs or upgrade discussions that I have come across have discussed it, so any push in the right direction would be helpful!

TimothyGu:
Those are just about the changes done. I'd love it if you could help prepare some documentation for using client mode @[pug-en](https://github.com/pugjs/pug-en). Just add a src/api/client.md and follow other src/api/*.md's syntax.

strand:
@TimothyGu It looks like the client API documentation still needs to be written in @pug-en. If I submit a draft PR, will you review and advise me on meeting pug's standards.

ForbesLindesay:
Yes, either @TimothyGu or me will review any pull request to the pug-en documentation.

matthewjumpsoffbuildings:
What is the status of this? The docs still seem out of date, Im not sure how to implement client side compilation currently...

Im using webpack, no idea how to get my JS files to reference the pug and correctly precompile the templates to functions etc

PR #2980, opened 2018-03-09
Added browser/amd/commonjs pug-runtime support

### LINKED ISSUE #2902: what version? what browser?
**Pug Version:** 2.0.0-rc4

**Node Version:** 9.1.0

image

niftylettuce:
This closes #2902, #2547, and #2407

### DIFF
--- a/packages/pug-runtime/index.js
+++ b/packages/pug-runtime/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var pug = {};
 var pug_has_own_property = Object.prototype.hasOwnProperty;
 
 /**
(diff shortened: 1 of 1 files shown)


niftylettuce:
cc @ForbesLindesay

ForbesLindesay:
I think I'd prefer to use https://github.com/ForbesLindesay/umd to autogenerate this from source.

niftylettuce:
@ForbesLindesay so all I'd need to do is change this:

```js
if (typeof module !== 'undefined' && typeof module.exports !== 'undefined') {
  module.exports = pug;
}
else {
  if (typeof define === 'function' && define.amd) {
    define([], function() {
      return pug;
    });
  }
  else {
    window.pug = pug;
  }
}
```

to

```js
return pug;
```

And then what would your preferred build script be?

PR #3002, opened 2018-04-17
[pug-filters] Throw error if error when attempting to require filter

damassi:
When an error is encountered in a `jstransformer` package that is being auto-imported, currently the exception is silently caught so that its hard to debug. For example

```javascript
// node_modules/jstransformer-hello-error/index.js

throw new Error('Oh no')

exports.render = () => {
  ...
}

// my-pug-template
.div
  :hello-error() 
```
Returns "cannot find `:hello-error` filter" rather than the exception being thrown somewhere in the source code.

### DIFF
--- a/packages/pug-filters/lib/run-filter.js
+++ b/packages/pug-filters/lib/run-filter.js
@@ -8,38 +8,34 @@ var resolve = require('resolve');
 module.exports = filter;
 function filter(name, str, options, currentDirectory, funcName) {
   funcName = funcName || 'render';
-  var tr;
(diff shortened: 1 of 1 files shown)


shesek:
This had the unintended side effect of making filters not work with browserified bundles. The `resolve` package is unable to find the jstransformer module when everything bundled into a single file, and `require.resolve()` doesn't exists in browserify. This would work before the change in this PR, if it just tried to `require()` it directly.

Any thoughts on how to best address this?

Issue #3040, opened 2018-07-30
Plugin preLex call


SergeShkurko:
**Pug Version:** 2.0.3

**Node Version:** 9.9.0

## My plugin code

```js
module.exports = {
        preLex(contents) {
            console.log('contents', contents)
            return contents
        },
        postLex(tokens) {
            console.log('tokens', tokens)
            return tokens
        }
}
```

## Example 1:

```js
pug.compile('b.foo fooText', {
  filename: path,
  pretty: true,
  plugins: [ MyOwnPlugin ],
});
```

#### Output from my plugin `console.log`:
- [ ] contents (?) 
- [x] tokens

## Additional Comments

Function  `preLex` not called if template compile from string

If file loaded via `include` function  `preLex`  called normally (see example 2)

## Example 2:
```pug
// file.pug
b.bar barText
```
```js
// index.js
pug.compile('include file.pug \n b.foo fooText', {
  filename: path,
  pretty: true,
  plugins: [ MyOwnPlugin ],
});
```

`preLex` called only for included file

#### Output from my plugin `console.log`:
- [ ] contents from string (?) 
- [x] tokens from string
- [x] contents from file
- [x] tokens from file

ForbesLindesay:
Thanks for the bug report, and the fix. This will be resolved in the next release.

PR #3041, opened 2018-07-30
Fix preLex fire support for plugins

SergeShkurko:
Function preLex not called if template compile from string (See #3040)


ForbesLindesay:
Thanks! I can't promise when I'll next get round to cutting a release.

SergeShkurko:
I will wait with impatience.

PR #3073, opened 2018-11-03
pug_attr should preserve value token boundaries even when escaped=false

mikesamuel:
# Before

```sh
$ node
> const pug_attr = require('pug-runtime').attr;
undefined
> pug_attr('name', 'val">', false)
' name="val">"'
```

# After

```sh
$ node
> const pug_attr = require('pug-runtime').attr;
undefined
> pug_attr('name', 'val">', false)
' name="val&#34;>"'
```


ForbesLindesay:
The entire point of unescaped mode is that it's totally unescaped. Creating a half escaped mode like this adds a lot of extra complexity and potential for mistakes and mis-understandings.

PR #3097, opened 2019-01-24
feat: exposed handleTemplateCache in module exports

niftylettuce:


ForbesLindesay:
Why is this needed, I'd prefer to keep this particular API private, as it's not been designed for public consumption. If we expose it, we need to avoid breaking it without major version releases, which is otherwise un-necessary.

niftylettuce:
This should be backwards compatible so only a minor bump would be needed.  I was going to use this in `cache-pug-templates`, but I actually am going with a different approach now for that.

PR #3098, opened 2019-01-25
Consume all whitespace and commas before the key

shreeve:
This fix is essentially be a no-op. It lets `pug` be a little more lenient about commas in the attributes list. Right now, commas between attributes are optional and essentially just ignored. This small patch lets you begin a list of attributes with a comma or put several in the list, without giving any complaints. It just happily skips over them as if they were whitespace.

**Why is this PR requested?**

Because OCD makes me itch when I see these attributes not line up:

untitled 2019-01-25 09-29-57

To line up the code, we've been using this:

image

But this is a little awkward too...

So, this PR allows valid `pug` (ie - a comma between attributes) to be used:

untitled 2019-01-25 09-39-51

Maybe this is also a little weird... but, the PR allows these extra commas to be happily ignored.

### DIFF
--- a/packages/pug-lexer/index.js
+++ b/packages/pug-lexer/index.js
@@ -1062,9 +1062,9 @@ Lexer.prototype = {
     var key = '';
     var i;
     
-    // consume all whitespace before the key
+    // consume all whitespace and commas before the key
     for(i = 0; i < str.length; i++){
-      if(!this.whitespaceRe.test(str[i])) break;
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
I don't think this leads to cleaner or clearer code. Thank you for your contribution though.

shreeve:
Yeah, I reverted the changes in the pug files also. Couldn't really find a nice way to "line up" the attrs, so just gave up and wrote up in the "normal" way. :-)

PR #3101, opened 2019-02-01
Add codeGen plugin support

SerayaEryn:
This adds the possibility to replace `pug-code-gen` with a plugin.

### DIFF
--- a/packages/pug/lib/index.js
+++ b/packages/pug/lib/index.js
@@ -50,16 +50,7 @@ function applyPlugins(value, options, plugins, name) {
   }, value);
 }
 
-function getPlugin(plugins, name, defaultFunction) {
-  for (var plugin of plugins) {
-    if (plugin[name]) {
-      return plugin[name];
-    }
-  }
-  return defaultFunction;
-}
-
-function findReplacementFunc(plugins, name) {
+function findReplacementFunc(plugins, name, defaultFunction) {
   var eligiblePlugins = plugins.filter(function (plugin) {
     return plugin[name];
   });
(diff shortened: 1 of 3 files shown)


packages/pug/lib/index.js
```diff
@@ -50,6 +50,15 @@ function applyPlugins(value, options, plugins, name) {
   }, value);
 }
 
+function getPlugin(plugins, name, defaultFunction) {
+  for (var plugin of plugins) {
+    if (plugin[name]) {
+      return plugin[name];
+    }
+  }
+  return defaultFunction;
```
ForbesLindesay: This function needs to throw an error if there is more than one plugin implementing the same `name`. i.e. this approach does not support multiple code-gen plugins.

SerayaEryn: I decided to remove the new function and reuse `findReplacementFunc()` since it was almost doing what I needed.

SerayaEryn:
That's totally fine. Thanks for the review. :)

packages/pug/lib/index.js
```diff
@@ -172,7 +172,7 @@ function compileBody(str, options){
 
   // Compile
   ast = applyPlugins(ast, options, plugins, 'preCodeGen');
-  var js = generateCode(ast, {
+  var js = findReplacementFunc(plugins, 'codeGen', generateCode)(ast, {
```
ForbesLindesay: If you made this:

```suggestion
  var js = (findReplacementFunc(plugins, 'codeGen') || generateCode)(ast, {
```

You wouldn't need any changes to `findReplacementFunc` at all.

SerayaEryn: I made the suggested change and removed the changes to `findReplacementFunc()`.

PR #3108, opened 2019-02-08
Update dev dependencies

SerayaEryn:
This updates the dev dependencies and adds Node.js 10 and 11 in the `.travis.yml`.


SerayaEryn:
I dropped node v4 from the `.travis.yml` since it is EOL since 30 April 2018 and it isn't supported by `jest`.

PR #3119, opened 2019-03-11
Add test for inline and include markdown-it

natmbake:


natmbake:
There is a bug in node v11.11.0, so to overcome it, set a fixed node version to travis.yml

PR #3141, opened 2019-05-14
bump constantinople to 3.1.2

pklingem:
see: https://snyk.io/vuln/npm:constantinople:20180421

### DIFF
--- a/packages/pug-code-gen/package.json
+++ b/packages/pug-code-gen/package.json
@@ -6,7 +6,7 @@
     "pug"
   ],
   "dependencies": {
-    "constantinople": "^3.0.1",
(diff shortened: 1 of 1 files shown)


pklingem:
@ForbesLindesay any chance to get this merged?  It's vulnerability that's coming up on our security scans.

ForbesLindesay:
You just need to update your yarn.lock/package-lock.json file - `^3.0.1` is a semver range that includes 3.1.2 so the fix is already included. Both npm and yarn always install the latest matching version unless you have a lock file.

pklingem:
thank you @ForbesLindesay

PR #3152, opened 2019-06-15
Allow spread operator in attributes

ezhlobo:
That is a tiny fix to have better supporting of spread operator in attributes.

For example the following case threw a SyntaxError:
```pug
div(anything ...props)
```

Now that works. It will help me to provide better experience for react users.


ezhlobo:
@pugjs/maintainers hey, that is a tiny pull request (-1 +5). The rest LOC changes are about tests (and they mostly autogenerated). I'll appreciate if you take a look at this, some tools which I use every day depends on it.

Issue #3170, opened 2019-08-13
`each` should recognize iteration protocol objects


strugee:
**Pug Version:** 2.0.4

**Node Version:** 8.10.0

## Input JavaScript Values

```js
var pug = require('pug');

var map = new Map([['foo', 'bar']]);

var html = pug.renderFile('input.pug', { some_iterable: map });

console.log(html);
```

## Input Pug

```pug
h1 Pug iteration test file
each el in some_iterable
     p= el[0]
     p= el[1]
```

## Expected HTML

```html
<h1>Pug iteration test file</h1><p>foo</p><p>bar</p>
```

## Actual HTML

```html
<h1>Pug iteration test file</h1>
```

## Additional Comments

Here's the relevant MDN documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

I would be happy to prepare a PR implementing this change but wanted to open an issue first to make sure there aren't any issues I haven't thought of yet. I also _think_ this will be backwards-compatible pretty easily, but don't have the time to double-check at the moment. I'll do that soon.

ForbesLindesay:
This will require us to adopt some new syntax. I would suggest something like `for item of iterable` since we already have `for` as a reserved word, and could easily enough detect the `in`/`of` distinction.

maxrumsey:
@strugee Are you currently working on a PR for this? Because if not I'd be happy to take it on.

ForbesLindesay:
@maxrumsey I've marked it as being worked on. Based on the assumption that one of you or @strugee will be working on it :)

strugee:
@maxrumsey I am not. If no one lands this I'll probably eventually prepare patches just because I want this in my own projects, but I'm not doing anything right now and would be thrilled if someone else implemented this for me :-)

maxrumsey:
Yeah I'd be happy to take a stab at this!

maxrumsey:
@ForbesLindesay Considering the size of Pug, do you have any tips for where I could get started? See the gitter chat.

ForbesLindesay:
You'll need to start by building an eachOf variant of https://github.com/pugjs/pug/blob/fa8d287ac7155e4e6dd56ff38f4495135659532a/packages/pug-lexer/index.js#L950-L972

You'll then need to handle it in the parser https://github.com/pugjs/pug/blob/fa8d287ac7155e4e6dd56ff38f4495135659532a/packages/pug-parser/index.js#L235-L236

[...]

PR #3179, opened 2019-09-11
Adding 'each .. of' syntax

maxrumsey:
This PR adds support for iterating over maps and other iterable objects with the JS syntax `each var of map`. This example patches issue #3170. The syntax `each var of map` and `for var of map` has been introduced, but `each` and `for` can be used interchangeably, like with the original each syntax. 

All tests run and pass, but I haven't added any tests except for a small fix to recognise the `eachOf` function is the `pug-lexer`.

### JS
```js
var pug = require('pug');

var map = new Map([['foo', 'bar'], ['a', 'b']]);

var html = pug.renderFile('code.pug', { some_iterable: map });

console.log(html);
```
### Pug
```
h1 Pug iteration test file
each el of some_iterable
     p= el[0]
     p= el[1]
```
### Output
```html
<h1>Pug iteration test file</h1><p>foo</p><p>bar</p><p>a</p><p>b</p>
```

### DIFF
--- a/packages/pug-code-gen/index.js
+++ b/packages/pug-code-gen/index.js
@@ -65,7 +65,6 @@ function Compiler(node, options) {
   this.mixins = {};
   this.dynamicMixins = false;
   this.eachCount = 0;
-  this.eachOfCount = 0;
   if (options.doctype) this.setDoctype(options.doctype);
   this.runtimeFunctionsUsed = [];
   this.inlineRuntimeFunctions = options.inlineRuntimeFunctions || false;
--- a/packages/pug-code-gen/index.js
+++ b/packages/pug-code-gen/index.js
@@ -767,6 +767,52 @@ Compiler.prototype = {
     this.buf.push('  }\n}).call(this);\n');
   },
 
+  visitEachOf: function(each){
+    var indexVarName = each.key || 'pug_index' + this.eachCount;
+    this.eachCount++;
+
+    this.buf.push(''
+      + '// iterate ' + each.obj + '\n'
+      + ';(function(){\n'
+      + '  var $$obj = ' + each.obj + ';\n'
+      + '  if (\'number\' == typeof $$obj.length) {');
+
+    if (each.alternate) {
+      this.buf.push('    if ($$obj.length) {');
+    }
+
+    this.buf.push(''
+      + '      for (var ' + indexVarName + ' = 0, $$l = $$obj.length; ' + indexVarName + ' < $$l; ' + indexVarName + '++) {\n'
+      + '        var ' + each.val + ' = $$obj[' + indexVarName + '];');
+
+    this.visit(each.block, each);
+
+    this.buf.push('      }');
+
+    if (each.alternate) {
+      this.buf.push('    } else {');
+      this.visit(each.alternate, each);
+      this.buf.push('    }');
+    }
+
+    this.buf.push(''
+      + '  } else {\n'
+      + '    var $$l = 0;\n'
+      + '    for (var ' + indexVarName + ' of $$obj) {\n'
+      + '      $$l++;\n'
+      + '      var ' + each.val + ' = ' + indexVarName + ';');
+
+    this.visit(each.block, each);
+
+    this.buf.push('    }');
+    if (each.alternate) {
+      this.buf.push('    if ($$l === 0) {');
+      this.visit(each.alternate, each);
+      this.buf.push('    }');
+    }
+    this.buf.push('  }\n}).call(this);\n');
+  },
+
   /**
    * Visit `attrs`.
    *
--- a/packages/pug-code-gen/index.js
+++ b/packages/pug-code-gen/index.js
@@ -769,34 +769,13 @@ Compiler.prototype = {
   },
 
   visitEachOf: function(each){
-    var valVarName = 'pug_val' + this.eachOfCount;
-    if (each.key) {
-      var keyVarName = 'pug_key' + this.eachOfCount
-    }
-    this.eachOfCount++;
-
     this.buf.push(''
       + '// iterate ' + each.obj + '\n'
-      + ';(function(){\n'
-      + '  var $$obj = ' + each.obj + ';\n'
-      + '  var $$l = 0;\n')
-    if (each.key) {
-      this.buf.push('  for (const [' + keyVarName + ', ' + valVarName + '] of $$obj) {\n');
-    } else {
-      this.buf.push('  for (const ' + valVarName + ' of $$obj) {\n');
-    }
-    this.buf.push('    $$l++;\n');
-    if (each.key) {
-      this.buf.push(''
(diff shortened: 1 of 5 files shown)


packages/pug-lexer/index.js
```diff
@@ -971,6 +971,31 @@ Lexer.prototype = {
     }
   },
 
+  /**
+   * EachOf.
+   */
+
+  eachOf: function() {
+    var captures;
+    if (captures = /^(?:each|for) +\[?([a-zA-Z_$][\w$]*)(?: *, *([a-zA-Z_$][\w$]*))?\]? * of *([^\n]+)/.exec(this.input)) {
```
ForbesLindesay: The way this is written allows unbalanced brackets: `each key, value] of list` as well as omiting the brackets `each key, value of list`. What I think we should have is just two possible forms, `each value of list` and `each [key, value] of list`. At the lexer/parser we can just dump the entire `[key, value]` string in the `value` property and then when we output it we'll just be treating it as JavaScript destructuring.

packages/pug-code-gen/index.js
```diff
@@ -767,6 +768,37 @@ Compiler.prototype = {
...
+    if (each.key) {
+      var keyVarName = 'pug_key' + this.eachOfCount
+    }
+    this.eachOfCount++;
+
+    this.buf.push(''
+      + '// iterate ' + each.obj + '\n'
+      + ';(function(){\n'
+      + '  var $$obj = ' + each.obj + ';\n'
+      + '  var $$l = 0;\n')
```
ForbesLindesay: I don't think `$$l` is read anywhere, can we get rid of it?

packages/pug-code-gen/index.js
```diff
@@ -767,6 +768,37 @@ Compiler.prototype = {
...
+    } else {
+      this.buf.push('  for (const ' + valVarName + ' of $$obj) {\n');
+    }
+    this.buf.push('    $$l++;\n');
+    if (each.key) {
+      this.buf.push(''
+        + '    var ' + each.key + ' = ' + valVarName + ';'
+        + '    var ' + each.val + ' = ' + keyVarName + ';');
+    } else {
+      this.buf.push('    var ' + each.val + ' = ' + valVarName + ';')
```
ForbesLindesay: Lets just use the requested variable names directly in place of `valVarName`

packages/pug-code-gen/index.js
```diff
@@ -767,6 +768,37 @@ Compiler.prototype = {
...
+    var valVarName = 'pug_val' + this.eachOfCount;
+    if (each.key) {
+      var keyVarName = 'pug_key' + this.eachOfCount
+    }
+    this.eachOfCount++;
+
+    this.buf.push(''
+      + '// iterate ' + each.obj + '\n'
+      + ';(function(){\n'
+      + '  var $$obj = ' + each.obj + ';\n'
```
ForbesLindesay: We should be able to inline `each.obj` in the js `for` statement. It only gets evaluated once anyway, so there's no need to assign it to a variable.

packages/pug-code-gen/index.js
```diff
@@ -767,6 +768,37 @@ Compiler.prototype = {
...
+      this.buf.push(''
+        + '    var ' + each.key + ' = ' + valVarName + ';'
+        + '    var ' + each.val + ' = ' + keyVarName + ';');
+    } else {
+      this.buf.push('    var ' + each.val + ' = ' + valVarName + ';')
+    }
+
+    this.visit(each.block, each);
+
+    this.buf.push('  }\n}).call(this);\n');
```
ForbesLindesay: If you make the changes proposed above, there will be no variables that are not block scoped to the loop anyway, so we can remove the self invoking function that wraps the loop.

maxrumsey:
@ForbesLindesay I've updated the parser as well as the code-gen library. Anything else needing a change?

packages/pug-lexer/index.js
```diff
@@ -971,6 +971,31 @@ Lexer.prototype = {
     }
   },
 
+  /**
+   * EachOf.
+   */
+
+  eachOf: function() {
+    var captures;
+    if (captures = /^(?:each|for) (.*) * of *([^\n]+)/.exec(this.input)) {
```
ForbesLindesay:
```suggestion
    if (captures = /^(?:each|for) (.*) of *([^\n]+)/.exec(this.input)) {
```

The `.*` will already consume all the whitespace, which I think is fine.

packages/pug-lexer/index.js
```diff
@@ -971,6 +971,31 @@ Lexer.prototype = {
...
+      this.consume(captures[0].length);
+      var tok = this.tok('eachOf', captures[1]);
+      tok.value = captures[1] || null;
```
ForbesLindesay: This should never be `null`.

```suggestion
      tok.value = captures[1];
```

ForbesLindesay: I think we should have something like:

```js
if (!(/^[a-zA-Z_$][\w$]*$/.test(tok.value.trim()) || /^\[ *[a-zA-Z_$][\w$]* *\, *[a-zA-Z_$][\w$]* *\]$/.test(tok.value.trim())) {
      this.error(
        'MALFORMED_EACH_OF_LVAL',
        'The value variable for each must either be a valid identifier (e.g. `item`) or a pair of identifiers in square brackets (e.g. `[key, value]`).'
      );
}
```

We can always update this to be more flexible later, if we are willing to actually validate that it's a valid JS destructuring pattern, but allowing everything now makes it difficult to safely add validation later.

packages/pug-parser/index.js
```diff
@@ -761,6 +763,20 @@ loop:
     return node;
   },
 
+  parseEachOf: function(){
+    var tok = this.expect('eachOf');
+    var node = {
+      type: 'EachOf',
+      obj: tok.code,
+      val: tok.val,
+      key: tok.key,
```
ForbesLindesay: There is no `key` now.

packages/pug-code-gen/index.js
```diff
@@ -65,6 +65,7 @@ function Compiler(node, options) {
   this.mixins = {};
   this.dynamicMixins = false;
   this.eachCount = 0;
+  this.eachOfCount = 0;
```
ForbesLindesay: This is no longer used for anything.

maxrumsey:
@ForbesLindesay That's awesome to hear. Anything else that could be fixed up? :)

ForbesLindesay:
This is now merged and will go in the next release.

maxrumsey:
That's awesome to hear. Thanks for the help with getting me started. I'll post the PR now.

nmggithub:
> This is now merged and will go in the next release.

Is it available now?

GHolk:
Accidentally found this feature. This feature is great, but does not mentioned in [pug's iteration docs](https://pugjs.org/language/iteration.html).

Issue #2918, opened 2017-12-06
Malformed each error


mahish:
**Pug Version:** "gulp-pug": "^3.3.0" which uses "pug": ">=2.0.0-alpha <3",

**Node Version:** v8.2.1

## Input Pug

```pug
mixin notificationSet(opts, notifications)
  - const headlineFontWeight = 'headline-' + opts.headline.fontWeight
  - const classes = headlineFontWeight
  - const versions = notifications.content
  - const items = notifications.items

  each content from versions
    .notification-subset(class= classes)
      each notification from items
        +notification(notification.theme, content)
```

## Error

```
Error: …/src/views/_mixins/notificationSet.pug:7:7
    5|   - const items = notifications.items
    6|
  > 7|   each content from versions
-------------^
    8|     .notification-subset(class= classes)
    9|       each notification from items
   10|         +notification(notification.theme, content)

malformed each
```

## Additional Comments
I havent found any single documentation on `malformed each` error. I tried to change variable names for something else but without any result. Am I missing something? I've never had problem with `each` or nested `each` before.

mahish:
lol, kill me pls. Wrote `from` instead of `in`.

margaritluch:
lmao had the same stupid error 🤣

PR #3185, opened 2019-09-18
feat: improve error message for malformed each

ForbesLindesay:
It now includes a description of what `each` should look like. We guide people towards the newer `each ... of ...` syntax, rather than the older `each ... in ...` syntax.


ForbesLindesay:
Hopefully this would make #2918 easier to resolve.

PR #3187, opened 2019-10-15
Fix package.json `repository` definition

strugee:
See https://docs.npmjs.com/files/package.json#repository. tl;dr: the `url` field here is supposed to be something that can be passed directly to Git. It cannot be something that will give you HTML.


ForbesLindesay:
The important thing is where the URL takes you when you click on the link to the repository on the npm website.

PR #3213, opened 2020-02-08
Allow filters to read non-text include files

brewingcode:
I have some custom filters that I use with Pug's `include`, but the filters were choking because the incoming "text" from the `include` was wrong: it was automatically encoded as `utf8` but my filters need `Buffer`.

So the simplest thing was to just expose a new option that, given the arguments to `load.read`, would simply tell Pug to either continue specifying `utf8` encoding, or just leave the encoding out and return the `Buffer`.

A very stripped-down example looks like this:

```js
// options.js
exports.skipEncoding = function(filename, options) { return filename.match(/png$/) };
exports.filters = {
  png: function(data, options) { /* do something heinous */ },
};
 ```

```pug
// my-template.pug
include:png image.png
```

I also noticed one broken test (unrelated to my changes) and fixed it, too.

This use-case is minor enough that I didn't even want to mention it in any docs.


brewingcode:
I originally had a commit in there (1ccbcfa) to fix what I thought were broken tests, but I guess my local setup is different from CI.

brewingcode:
Thanks for the reply! I was going for the absolute smallest change to accomplish what I needed.

>  a cleaner approach to this would be to do it as part of the filter

Okay, sounds reasonable. In order to add a magic key like `renderBuffer` to a filter, I'm looking here:

https://github.com/pugjs/pug/blob/bb0731f75813aa30d8e077808b5465a67ef284ef/packages/pug-filters/lib/handle-filters.js#L58

At this point, the file has already been read by `pug-load`, so are you suggesting that I re-read the file with `fs` in this function? I'm generally not a performance freak, but even I would think twice about reading the same file over and over (each use of the filter in my .pug template runs this `filterFileWithFallback` function, right?).

I also don't know all the other cases where a filter assumes its input is plain text, this `filterFileWithFallback` function is simply the first place pug barfed when I tried passing `options: { filters: { png: { renderBuffer: ...`.

I guess what I'm asking is: could you give me a bit more guidance on how to safely add a `renderBuffer` check to the filtering module? I made my change at `load.read()` because it was the one single place where the files are read, AND only users who passed `options.skipEncoding` would ever be impacted by the change (so risk of breaking anybody else is zero). Changing a core part of pug was a bit more than I was willing to do.

brewingcode:
Rebased against latest master (bb0731f75813aa30d8e077808b5465a67ef284ef), previously change was against 201b9c78431e172dc90a6d258af982e2d1191da8

ForbesLindesay:
I think the process should be:

1. change `load.read` to return a Buffer (https://github.com/pugjs/pug/blob/master/packages/pug-load/index.js#L83)
2. `load.file` can then just call `.toString('utf8)` on the buffer
3. If it's not a raw include with filters, we can call `.toString('utf8')` on https://github.com/pugjs/pug/blob/master/packages/pug-load/index.js#L28
4. in pug-filters, we can call `.toString('utf8')` before calling any filters that expect a string.

This will be a breaking change for pug-load, but should be non-breaking for everything else.

brewingcode:
> This will be a breaking change for pug-load, but should be non-breaking for everything else.

I mean, right now as a new option it's non-breaking to everyone. :P But yes, I do see the value in having this be a default, rather than some buried option.

I'll see if I can manage it by changing `pug-load` and `pug-filters` as per your steps.

brewingcode:
OK, here's another attempt, slightly modified from your steps.

At first I tried your steps as-is, but step #3 (where I probably totally botched modifying https://github.com/pugjs/pug/blob/master/packages/pug-load/index.js#L28) completely exploded the tests with 55 failures, so I got spooked and went back to trying to change as few things as possible.

So I added a second property to `node.file` called `node.file.raw` (to live next to to `node.file.str`), and in `pug-filters` I use this new property if and only if the filter has the `renderBuffer` property. I do believe does what I need in a non-breaking way, as well.

packages/pug-load/test/__snapshots__/index.test.js.snap
```diff
@@ -122,6 +154,43 @@ Object {
...
+                      114,
+                      108,
+                      100,
+                      33,
+                      39,
+                      41,
+                      59,
+                      10,
+                    ],
+                    "type": "Buffer",
```
ForbesLindesay, line 192: You can customise how jest renders certain values in the snapshot via serializers. You can see an example in: https://github.com/pugjs/pug/blob/master/scripts/filename-serializer.js

I suggest we add one for `Buffer`. For now it could just call `.toString('utf8')` and tag the result as a Buffer. If we can have it detect Buffers that are probably binary data rather than text data, we could also have it output those as something like `Buffer(SHA_HASH)` so they're shorter and easier to review in the snapshots.

trasherdk:
Please educate me.
Are you proposing to expose clients to binary data? Or are there something else going on here?

packages/pug-load/test/__snapshots__/index.test.js.snap
brewingcode: Ah, that's nice. I was wondering that, but figured for the small buffers that existed, it wasn't too bad.

brewingcode: 425f44a

brewingcode:
@trasherdk Yes, but only via a specially-crafted filter. No current pug users are going to start seeing `Buffer`s getting barfed into HTML.

Here's a small example:

```js
// options.js
exports.filters = {
  png: {                                                        // instead of a function
    renderBuffer: function(buffer, options) {                   // specify a "renderBuffer" property 
      var data = Buffer.from(buffer).toString('base64');        // and THIS function recieves a buffer instead of text
      return '<img src="data:image/png;base64, ' + data + '"';
    }
  }
};
```

```pug
// foo.pug
include:png my-small-image.png
```

brewingcode:
@ForbesLindesay thanks! I made the comment scan a little more naturally in the example options.js notes for the Rolling Version and hit "Save Changes". I also edited [my comment above](https://github.com/pugjs/pug/pull/3213#issuecomment-624746062) to match.

If there's anything else, lemme know!

PR #3223, opened 2020-03-20
Update acorn and other dependencies

Noctfield:
* `with` -> 6.0.0
* `uglify-js` -> 2.8.29
* `mkdirp` -> 1.0.3
* `acorn-walk` + 7.1.1

Since [this advisory](https://github.com/advisories/GHSA-7fhm-mqm4-2wp7) disclosed a vulnerability in `acorn`, it and several dependencies depending on it should be updated.

`uglify-js2` has been merged into the NPM package, so the NPM version is used. In fact, it already has gone up to 3.x, and we might want to update there too.

Also, `acorn/dist/walk` had been put into a separate package, `acorn-walk`, so it's used instead.

Also `mkdirp`.

----

A regression occurred:

```
Summary of all failing tests
 FAIL  packages/pug/test/run-es2015.test.js
  ● test cases for ECMAScript 2015 › attr                                                                                 
SyntaxError: Unexpected character '`'

      at JS_Parse_Error.get (eval at Object.<anonymous> (node_modules/uglify-js/tools/node.js:27:1), <anonymous>:84:23)
```

I'm not sure why it happens, since template string notation should be available long ago.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ npm-debug.log
 .release.json
 lerna-debug.log
 package-lock.json
+**/yarn-error.log
--- a/packages/pug-code-gen/package.json
+++ b/packages/pug-code-gen/package.json
@@ -13,7 +13,7 @@
     "pug-error": "^1.3.3",
     "pug-runtime": "^2.0.5",
     "void-elements": "^3.1.0",
-    "with": "^5.0.0"
+    "with": "^6.0.0"
   },
   "files": [
     "index.js"
--- a/packages/pug-lexer/package.json
+++ b/packages/pug-lexer/package.json
@@ -6,6 +6,7 @@
     "pug"
   ],
   "dependencies": {
+    "acorn-walk": "^7.1.1",
     "character-parser": "^2.1.1",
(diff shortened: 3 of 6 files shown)


ricksterhd123:
@Noctfield 

Having looked around a bit it seems that uglifyjs is causing this error in https://github.com/mishoo/UglifyJS2/blob/b39228892d9737421027ce2e9d944976ffaee654/lib/parse.js sorry if that's obvious. I've run `npm ls acorn` in my package that uses pug and it comes up with this:
image

I'm not sure why this error has occurred it just doesn't seem to provide much detail, but this dependency ought to be updated asap. I hope this helps. Thanks for trying.

ricksterhd123:
According to https://github.com/mishoo/UglifyJS2#note and from mishoo/UglifyJS2#3829:
uglify-js only supports JavaScript (ECMAScript 5)
To minify ECMAScript 2015 or above, transpile using tools like Babel.

ForbesLindesay:
Thanks for the contribution. This has now been done via another PR.

Noctfield:
Nice work 😘 Glad to see it taken care of.

Issue #3116, opened 2019-02-21
Including file without extension as Plain Text


dartess:
**Pug Version:** 2.0.3

**Node Version:** 10.15.0

## Input Pug

```pug
include ./.browserslistrc
```

## Expected HTML

```html
-- raw text from non-Pug file ".browserslistrc" --
```

## Additional Comments
Pug tries to insert a file without extension as a pug file, which is why you cannot insert a file without extension as plain text.

PR #3228, opened 2020-04-05
Hotfix/inlcude files without ext

dartess:
Hello everyone here!

I am trying to solve the problem I found with including files without extension as raw https://github.com/pugjs/pug/issues/3116 .

The error occurs at the parsing stage ( https://github.com/pugjs/pug/blob/master/packages/pug/lib/index.js#L104 ), when `path.extname(token.val)` says that the file has no extension, so pug add `.pug` extension to the file.

This is true for most cases, but false for files without an extension.

The only way to fix this at the parsing stage is to check for a file on the file system. But I think it will be too difficult operation for the parsing stage. Therefore, I tried to solve the problem on the other hand, at the stage of reading the file.

If we determine that the file with the type `Include` could not be read, we try to change its type to `RawInclude`, delete the extension and read it again.

Since the type was initially defined as `Include`, and the type `RawInclude` requires `filters` field in `node`, I tried to assign filters always, regardless of the type. I did not understand the meaning of these filters and am not sure if this is the right solution. Perhaps it would be more correct to assign an empty array instead of filters at the time of type change.

If such a solution is acceptable at all, there is an open question:

Should I assign filters to a node regardless of its type and add corrections for tests? Or should I assign an empty array as filter at the time the node type changes?

### DIFF
--- a/packages/pug-load/index.js
+++ b/packages/pug-load/index.js
@@ -5,6 +5,22 @@ var path = require('path');
 var walk = require('pug-walk');
 var assign = require('object-assign');
 
+function tryReadFile(options, file, node) {
+  var pathSrt, str;
+  try {
+    pathSrt = options.resolve(file.path, file.filename, options);
+    file.fullPath = pathSrt;
+    str = options.read(pathSrt, options);
+  } catch (ex) {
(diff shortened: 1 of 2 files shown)


dartess:
did not receive feedback

PR #3242, opened 2020-05-01
Add Nodejs 14 to the Travis Builds

YasharF:


YasharF:
It might be time to drop support for Nodejs 6 and 8 as they are both in end-of-life.  Let me know if you would like me to remove them from the Travis configs as well, and to update readme or other documentation accordingly.

ForbesLindesay:
Thanks, I'm going to do this, but I'm going to take the opportunity to migrate to GitHub actions at the same time.

PR #3245, opened 2020-05-02
chore: use minimal settings to format test files

zackschuster:


zackschuster:
failure is due to missing the commit from #3247

PR #3259, opened 2020-05-25
chore: update with

### LINKED ISSUE #3239: projects using pug cannot be installed globally
**Pug Version:** 2.0.4

**Node Version:** 10.19

## Minimal steps to reproduce:

```
npm init -y
# change name to something unique; I used [@zingle/npm-error](https://www.npmjs.com/package/@zingle/npm-error) as a minimal example
npm publish --access all
# then switch to the root user and run:
npm install -g @zingle/npm-error      # ... or whatever package you named it
```

The npm command will exit with the following error:

```
npm WARN deprecated core-js@2.6.11: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.

> core-js@2.6.11 postinstall /usr/local/lib/node_modules/@zingle/npm-error/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"
```

## Additional Comments

This makes pug completely useless for a number of reasonable use cases.

ForbesLindesay:
closes #3239

### DIFF
--- a/packages/pug-code-gen/package.json
+++ b/packages/pug-code-gen/package.json
@@ -13,7 +13,7 @@
     "pug-error": "^1.3.3",
     "pug-runtime": "^2.0.5",
     "void-elements": "^3.1.0",
-    "with": "^6.0.0"
+    "with": "^7.0.0"
   },
   "files": [
     "index.js"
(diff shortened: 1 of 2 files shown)


IanVS:
I believe this should have been a breaking change, since `with@7` no longer supports node 8, which emans that `pug-code-gen` also stopped supporting node 8 starting in version 3.0.1.  I believe the right thing to do is to revert this change, and release a `4.0.0` version with this update to `with`.

PR #3262, opened 2020-05-26
Fix TypeScript and add eachOf token definition

Shinigami92:


Shinigami92:
Sadly the diff is not well readable :(
I added index.d.ts to files so it will also be published next time with npm

Also I updated the d.ts so it includes the new v5 eachOf token

PR #3269, opened 2020-06-18
Properly handle errors originating from included files when compileDebug is enabled

tmont:
When calling render/compile with `{ compileDebug: true }`, and an error was thrown from (or due to) a local, the `str` parameter in `pug_rethrow` was a `toJSON()`'d `Buffer`, and the rethrow code tried to `split` it like a string.

`str` would be something like:
```javascript
{
  type: 'Buffer',
  data: [ /* a bunch of numbers */ ]
}
```

which is what happens when you do something like `Buffer.from('foo').toJSON()`.

I couldn't really figure out where or how it happens, but I just made the rethrow function gracefully handle those serialized `Buffer` objects properly.

I wrote a test that failed (basically just calls an undefined function from an included file):
pug-error-reporting

and then fixed it in this PR.

### DIFF
--- a/packages/pug-runtime/index.js
+++ b/packages/pug-runtime/index.js
@@ -247,18 +247,24 @@ function pug_rethrow(err, filename, lineno, str) {
     err.message += ' on line ' + lineno;
     throw err;
   }
+  var context, lines, start, end;
   try {
-    str = str || require('fs').readFileSync(filename, 'utf8');
+    var encoding = 'utf8';
+    str = str || require('fs').readFileSync(filename, {encoding: encoding});
+    if (str.type === 'Buffer') {
+      str = Buffer.from(str.data).toString(encoding);
+    }
+    context = 3;
+    lines = str.split('\n');
+    start = Math.max(lineno - context, 0);
+    end = Math.min(lines.length, lineno + context);
   } catch (ex) {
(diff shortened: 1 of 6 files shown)


ForbesLindesay:
Thanks for this. The fix will need to be done where we create the `debug_sources` object (https://github.com/pugjs/pug/blob/06baa525a23049756de9587461d389a12bc12537/packages/pug/lib/index.js#L167)

I think the best bet would be to move that code into the `lex` method (https://github.com/pugjs/pug/blob/06baa525a23049756de9587461d389a12bc12537/packages/pug/lib/index.js#L85) so we can store the actual string, instead of a Buffer. The secondary advantage is that we only need debug_sources for files that we `lex` so it would produce a smaller output via not having any non-pug includes.

P.S. the filename should be available as `options.filename` in the `lex` function.

tmont:
@ForbesLindesay Finally had time to revisit this. I also got tired of manually modifying the sources to apply this patch every time my template code has an error just so I can see what it was. :)

I moved the Buffer check code to the `read` function in main `pug` package and out of `pug-runtime` (although I left in some better error reporting in the case of weird stuff happening when trying to manage the strings, but that should only happen in internal code I think).

ForbesLindesay (approved):

PR #3314, opened 2021-02-28
fix: sanitise and escape the `pretty` option

ForbesLindesay:


Nixinova:
Could this also be released as a 2.X patch so people with `"pug":"^2"` can receive it?

bramkragten:
> Could this also be released as a 2.X patch so people with `"pug":"^2"` can receive it?

You can just upgrade `pug-code-gen` to 2.0.3

Songkeys:
@bramkragten 

> You can just upgrade `pug-code-gen` to 2.0.3

But `pug` < 3.0.1 is labelled as a vulnerability. My security system keeps arguing that I should upgrade my `pug@^2` to `3.0.1`.

I think we should have a branch to cut a release for 2.X.

Issue #3263, opened 2020-06-03
each fails with variable named "officers"


jabberwock31415:
**Pug Version:** your version number here
3.0.0

**Node Version:** your version number here
10.21.0

## Input JavaScript Values

```js
router.get('/foo', (req, res, next) => {
    res.render('foo', {});
});
```

## Input Pug

```pug
- let officers = [{name:'alan'}, {name:'betty'}, {name:'charlie'}];
each dude in officers
  p #{dude.name}
```

## Expected HTML

```html
<p>alan</p>
<p>betty</p>
<p>charlie</p>
```

## Actual HTML

none--pug errors out

## Additional Comments

I get an error thrown from pug:

Error: /home/NPL/www/express-web/views/directory.pug:3:22
    1| - let officers = [{name:'alan'}, {name:'betty'}, {name:'charlie'}];
    2|
  > 3| each dude in officers
----------------------------^
    4|   p #{dude.name}
    5|

The value variable for each must either be a valid identifier (e.g. `item`) or a pair of identifiers in square brackets (e.g. `[key, value]`).

If I change the name of the variable from "officers" to something else ("ooficers" for example) the script works

jabberwock31415:
I get the same error with the new syntax:

```
Error: /mnt/storage/npl/www/express-web/views/foo.pug:3:22
    1| - let officers = [{name:'alan'}, {name:'betty'}, {name:'charlie'}];
    2| 
  > 3| each dude of officers
----------------------------^
    4|   p #{dude.name}
    5| 

The value variable for each must either be a valid identifier (e.g. `item`) or a pair of identifiers in square brackets (e.g. `[key, value]`).
```

error418:
Proposed a Pull-Request. Rename your variable to start with something other than `of`, in case you need a quick workaround

efiand:
Variable named `offers` gives the same error. Maybe, `of` sequence is a reason of this bug.

error418:
The variable must not start with any of the keywords (like `of`)

Delagen:
Half year past, breaking issue does not solved

trasherdk:
@Delagen Seems just fine to me. Did you try?

```javascript

- let officers = [{name:'alan'}, {name:'betty'}, {name:'charlie'}];

each officer in officers
  officer #{officer.name}

```

![image](https://user-images.githubusercontent.com/5003891/105269199-0e823a80-5bc6-11eb-8064-29470bc01666.png)

[...]

PR #3318, opened 2021-03-03
Fix to address #3263 for "in" usage

thekevinhunt:
I was still hitting issue #3263 when using "in" instead of "of" (which caused other odd problems).  The regex was looking for * (0+) spaces instead of + (at least one) space.  Using + instead prevents the regex from catching "in ofxxxxx".


Delagen:
@ForbesLindesay Please see it MR

PR #3373, opened 2022-05-11
[pug-lexer] Tailwind classes support

FlameArt:
Support for Tailwind classes syntax like `.lg:w-[555px]`, `.hover:border` etc
Test added

The only confilct I saw: colon `:` is used for block expansion, but only with space symbol between a blocks, Tailwind no need spaces in class names, so i just add negative lookahead for colon-space combination in regexp

### DIFF
--- a/packages/pug-lexer/index.js
+++ b/packages/pug-lexer/index.js
@@ -463,7 +463,7 @@ Lexer.prototype = {
    */
 
   className: function() {
-    var tok = this.scan(/^\.([_a-z0-9\-]*[_a-z][_a-z0-9\-]*)/i, 'class');
+    var tok = this.scan(/^\.([_a-z0-9\-]*[_a-z]([_a-z0-9\-\[\]]|:(?! ))*)/i, 'class');
     if (tok) {
       this.tokens.push(tok);
       this.incrementColumn(tok.val.length);
--- a/packages/pug-lexer/test/cases/classes.pug
+++ b/packages/pug-lexer/test/cases/classes.pug
@@ -12,3 +12,5 @@ a(class={foo: true, bar: false, baz: true})
 
 a.-foo
 a.3foo
+
+a.focus:border.lg:w-[555px]


byre54:
is this work already ? I uses Tailwind with Pug FYI, and it's kinda nice to have this.

webdiscus:
@FlameArt 

you can use Tailwind classes in `class` attribute:
```pug
//- instead of this non-standard syntax
a.focus:border.lg:w-[555px]
//- can be used clear standard syntax
a(class="focus:border lg:w-[555px]")
```

P.S.: Please don't pollute Pug with tailwind's non-standard syntax ;-)
Please note that several IDE linters give display errors and this syntax will not be supported in IDEs.

Shinigami92:
> @FlameArt
> 
> you can use Tailwind classes in `class` attribute:
> 
> ```
> //- instead of this non-standard syntax
> a.focus:border.lg:w-[555px]
> //- can be used clear standard syntax
> a(class="focus:border lg:w-[555px]")
> ```
> 
> P.S.: Please don't pollute Pug with tailwind's non-standard syntax ;-) Please note that several IDE linters give display errors and this syntax will not be supported in IDEs.

I think you see it from the wrong perspective :thinking: 
It's not supported, because it was not added / handled by pug
But this repo IS pug and this PR adds support for it!
Linter, formatter and syntax highlighter would need to be updated accordingly to support this new handling of classes after this PR gets released

kaceo:
> > P.S.: Please don't pollute Pug with tailwind's non-standard syntax > 
> I think you see it from the wrong perspective 🤔 

Hear hear.  Pug has not progressed feature-wise in past year but CSS frameworks like Tailwindcss has.  

Pug needs to evolve:
- learn to support extended class names including symbols like colon, slash, percent.  
- It needs to allow tags that needs many lines of class literals.  
- It needs to allow multiple attributes with same names not as errors, just warnings.  
- It needs a flexible multi-line quoting strategy for strings.

orenmizr:
~~status on this ? 
by a pug/vue user :/~~

I had to let it go as jade (pug) became more and more esoteric.

jpg:
Tailwindcss + Pug should be a match made in heaven; we just need pug to allow colons & slashes 🙏

I know we can put tailwind utility classes in the class attribute (I use the responsive utilities on every page, like lg:), but that undoes the syntactic beauty that brought us to pug in the first place

PR #3385, opened 2022-07-21
Attribute syntax

j0pgrm:
small change for &attribute alternative, not sure if it works.


ForbesLindesay:
This would be a breaking change, and the syntax for `&attributes` is intentionally verbose to make it clearer what it does and why it may be unsuitable/dangerous. A better option may be to support the js style `...` operator.

PR #3459, opened 2025-04-28
Fix build status badge

spenserblack:
I noticed that the badge was reporting issue https://github.com/badges/shields/issues/8671.

~~TBH I'm not sure right now why the badge is reporting "no status", but I'm assuming it's because the last run was quite a while ago.~~ Looks like it reports "no status" because the workflow only runs on *pull requests targeting `master`,* not on `master` itself. Does it make sense to remove the branch specifier for the badge, in that case? Or the workflow can run on pushes to `master`.


noraj (approved):

lbesecker195 (approved):
Verified: `.github/workflows/test.yml` exists, and the old `workflow/status` shields endpoint was retired in favor of `actions/workflow/status`. LGTM.

PR #3470, opened 2026-04-29
fix: guard AST node property lookups against prototype pollution (fixes #3414)

### LINKED ISSUE #3414: vulnerable undefined property lookup that escalating prototype pollution to remote code execution

Ayoub-glitsh:
## Problem

Several packages accessed properties on AST node objects without verifying they were own properties. In JavaScript, `obj.prop` traverses the prototype chain — so if an attacker has already achieved prototype pollution elsewhere in the app, they can inject values like `Object.prototype.block`, `.type`, `.val`, `.code`, `.line` and escalate to RCE during template compilation.

The three PoC gadgets from #3414 all exploit this pattern:
- **Gadget 1**: `Object.prototype.block` with a malicious `Text` node containing arbitrary JS in `line`
- **Gadget 2**: `Object.prototype.code` with a malicious `val`
- **Gadget 3**: `Object.prototype.val` injecting code into attribute generation

## Fix

Guard all property accesses that influence control flow or code generation with `Object.prototype.hasOwnProperty.call(node, 'prop')` before reading the value.

## Files changed

- **`pug-walk/index.js`** — the most critical, used by all other packages. `switch (ast.type)` and all `if (ast.block/alternate/consequent)` checks are now guarded.
- **`pug-code-gen/index.js`** — `node.line`, `node.block`, `tag.code`, `node.isInline`, `node.val`, `node.type` checks guarded.
- **`pug-linker/index.js`** — `node.type`, `node.call`, `node.mode`, `node.block`, `node.ignore`, `node.parents`, `node.textOnly` checks guarded.
- **`pug-filters/lib/handle-filters.js`** — `node.type`, `node.filters`, `node.block` checks guarded.

## Tests

648/650 tests pass. The 2 failures are pre-existing and unrelated — they test for old Node.js error message wording (`Cannot read property` vs the newer `Cannot read properties`) introduced in Node 16+.


Ayoub-glitsh:
@rollingversions bump

## pug-walk

### security

- Guard AST node property lookups with `hasOwnProperty` to prevent prototype pollution gadgets from escalating to RCE (fixes #3414)

## pug-code-gen

### security

- Guard AST node property lookups with `hasOwnProperty` to prevent prototype pollution gadgets from escalating to RCE (fixes #3414)

## pug-linker

### security

- Guard AST node property lookups with `hasOwnProperty` to prevent prototype pollution gadgets from escalating to RCE (fixes #3414)

## pug-filters

### security

- Guard AST node property lookups with `hasOwnProperty` to prevent prototype pollution gadgets from escalating to RCE (fixes #3414)

